PR #2, opened 2016-03-06
depend: update to upstream bitflags release

cardoe:
Upstream bitflags now supports no_std as a feature so switch to that.

Signed-off-by: Doug Goldstein <cardoe@cardoe.com>


phil-opp:
Thanks a lot!

PR #5, opened 2016-06-05
Add bootloader name tag

Ketsuban:
This was low-hanging fruit, since the tag has a simple structure and the payload is specified to be valid UTF-8.


phil-opp:
Looks great, thanks a lot!

PR #8, opened 2016-09-09
Whitespace fixes and multiboot2 module support

### LINKED ISSUE #7: No multiboot module support
Hey, thanks for this crate.
I'm willing to implement this, but if you don't want it in your repository I can do it in my own fork.

4e554c4c:
Fixes #7 
I think I'd also like to add a ModuleIter but for that I'd have to change some other things with TagIter and such. Sorry for taking so long but it took me awhile to get a module working that I could test with.

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -37,11 +37,11 @@ impl Iterator for ElfSectionIter {
             let next_section_addr = (self.current_section as *const _ as u32) + self.entry_size;
             self.current_section = unsafe{ &*(next_section_addr as *const ElfSection) };
             self.remaining_sections -= 1;
-			if section.typ == ElfSectionType::Unused as u32 {
-				self.next()
-			} else {
-	            Some(section)
-			}
+            if section.typ == ElfSectionType::Unused as u32 {
+                self.next()
+            } else {
+                Some(section)
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


phil-opp:
Thank you very much!

4e554c4c:
@phil-opp, how would you suggest I go about implementing a ModuleIter?

phil-opp:
What's the problem with the current `TagIter`?

The problem is that people depend directly on this git repo, since I never pushed it to crates.io. I just fixed this (`multiboot2` is now on crates.io) and I try to update the blog_os post in the next days. However I still think we should avoid breaking changes in the near term.

4e554c4c:
The only problem that I have with `TagIter` is that there it is possible to get only one Tag of each type. There can be  multiple tags of the type module, and it is only possible to get the first one at the moment. This can be solved by either making `TagIter` public and having public methods on it, or create a `ModuleIter` struct.

phil-opp:
Ah, thanks for the explanation. Are there any other tags with this property (multiple tags with same type)? If the module tag is the only tag with multiple copies, I would prefer the `ModuleIter` option.

In either case, I think we should rename the `module_tag` function to `module_tags` and return some kind of `Iterator<&'static ModuleTag>` from it.

PR #9, opened 2016-09-15
Added ModuleIter, .gitignore change

4e554c4c:
Why I was originally wary about creating ModuleIter is because I
have to make Tag and TagIter public so that it can work. I can
change this (by practically rewriting TagIter in ModuleIter), but
I think that this works fine (issue/comment if you feel otherwise).

I also made a very small change to .gitignore so that I would not
accidentally commit my .swp files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+*.swp
--- a/src/header.rs
+++ b/src/header.rs
@@ -1,0 +1,30 @@
+#[repr(C)]
+pub struct Tag {
+    pub typ: u32,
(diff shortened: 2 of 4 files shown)


phil-opp:
Hmm, I would prefer to avoid new public items. AFAIK we could work around the “private type in public interface” error by creating a new (private) submodule for `Tag` and `TagIter`. Then we can make `Tag` and `TagIter` `pub` without exporting them from the crate.

4e554c4c:
@phil-opp Would this work?

phil-opp:
Looks good to me. Thanks a lot!

PR #12, opened 2016-09-29
Fix #11:  BootLoaderNameTag returning null char

### LINKED ISSUE #11: Returned strings include null terminator?
If you e.g. `writeln!(serial, "{:?}", info.boot_loader_name_tag().unwrap().name())`, you get `"GRUB 2.02~beta3\u{0}"` instead of just `"GRUB 2.02~beta3"`. Is this incorrect behavior? (It's unexpected, at least).

4e554c4c:
This is the same method I use to get the name in
`src/module.rs`, now neither return null characters.

### DIFF
--- a/src/boot_loader_name.rs
+++ b/src/boot_loader_name.rs
@@ -9,6 +9,11 @@ pub struct BootLoaderNameTag {
 
 impl BootLoaderNameTag {
     pub fn name(&self) -> &str {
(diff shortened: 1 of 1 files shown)


phil-opp:
Wow, that was quick! :) Thanks a lot!

4e554c4c:
No problem! I already had the solution in my module, so I just moved it over :)

PR #13, opened 2016-10-08
Replace overflowing cast in src/elf_sections.rs

4e554c4c:
When using a higher half kernel the cast to u32 will tuncate the
section address which will return an incorrect elf section reference
to the caller. This is fixed by casting instead to a u64.


phil-opp:
Thanks!

PR #14, opened 2016-10-08
Added name method to ElfSection

4e554c4c:


4e554c4c:
This commit introduces a problem with higher half kernels (such as mine) because it has to dereference a flat protected mode address to get the string table, instead of just giving that address to the caller like most do. A solution to this would be to add a closure in the arguments that maps a 32bit flat address to a 64bit correct address, but I'm not sure if this would make it too complicated to call.

src/elf_sections.rs
```diff
@@ -62,6 +62,29 @@ pub struct ElfSection {
...
+        unsafe {
+            let mut strlen = 0;
+            for i in 0.. {
+                if *name_byte.offset(i) == 0 {
+                    strlen = i as usize;
+                    break;
+                }
+            }
+
+            str::from_utf8_unchecked(
```
phil-opp: Maybe we should use the safe [from_utf8](https://doc.rust-lang.org/core/str/fn.from_utf8.html) instead? I'm not sure if the specification guarantees that the string is valid utf8… And even if it does, I think that the additional check is a good idea, because performance shouldn't really matter in this case.

phil-opp:
Regarding higher half kernels:

I'm not quite sure why this is a problem… All the pointers are relative to the multiboot header type, so it shouldn't be a problem to map it to a different address. Or am I missing something here?

4e554c4c:
@phil-opp The `strtabs.addr` field is, to my knowlege, a entirely arbitrary address so dereferencing it when the multiboot area is not identity mapped could lead to problems. All the other addresses are relative.

phil-opp:
Oh, I missed that.

Hmm… An alternative solution could be to introduce a new `StringTable` type:

```rust
impl ElfSectionsTag {
    pub fn string_table(&self) -> &StringTable {
        unsafe {
            let string_table_ptr = (&tag.first_section as *const ElfSection).offset(tag.shndx as isize);
             &*(string_table as *const StringTable)
         }
    }
}

struct StringTable([u8]);

impl StringTable {
    pub fn section_name(&self, section: &ElfSection) -> &'static str {
        let name_byte = self.0[section.name as usize];
        // ...
        unimplemented!();
    }
}
```

This way, you could adjust the reference returned by `string_table` before you derefence it.

4e554c4c:
@phil-opp I think that it would be better to not use a DST for the string table as the elf64 specification does not have a specific size for it. Also an address cannot be casted to a DST pointer, as it also contains a length. The string table struct could possibly contain a byte pointer to the beginning of the string table, but without exposing this field to the caller it would defeat the purpose of having another struct.

phil-opp:
Hmm.. How about a `struct StringTable(u8)`?

4e554c4c:
This uses three small unsafe blocks instead of one large one, I'm not convinced that it is better though. If you have a better idea I'm happy to change it.

src/elf_sections.rs
```diff
@@ -18,6 +18,40 @@ impl ElfSectionsTag {
...
+            &*((*string_table_ptr).addr as *const StringTable)
+        }
+    }
+}
+
+pub struct StringTable(u8);
+
+impl StringTable {
+    pub fn section_name(&self, section: &ElfSection) -> &'static str {
+        unsafe fn strlen(start: *const u8) -> usize {
```
phil-opp: Only `*start.offset(i)` is unsafe, so we could just do:

```rust
for i in 0.. {
    if unsafe { *start.offset(i) } == 0 {
        return i as usize;
    }
}
unreachable!()
```

This way it's much easier to spot the unsafe operations. On the other hand, I'm not comfortable with removing the `unsafe` from the `strlen` function (since it's not safe).

So I would prefer to remove the `strlen` function. Instead, we could do it like in your first commit:

```rust
let strlen;
for i in 0.. {
    if unsafe { *start.offset(i) } == 0 {
        strlen = i as usize;
        break;
    }
};
```

Or alternatively with a `while` loop:

```rust
let strlen = {
    let mut len = 0;
    while unsafe { *start.offset(len) } != 0 {
        len += 1;
    }
    len as usize
};
```

phil-opp:
Thanks a lot! :tada:

phil-opp:
Pushed to crates.io as version `0.2.1`

4e554c4c:
No problem! It's nice to have another release. Any other features needed for the near future?

phil-opp:
None that I can think of…

src/elf_sections.rs
bjorn3: from_utf8_lossy?

phil-opp: Unfortunately String and Cow aren't available, because this crate is `no_std`.

PR #15, opened 2016-10-10
Prevent overflow in `src/memory_map.rs`

4e554c4c:
This commit makes it so that pointers in `src/memory_map.rs` are casted only to other pointers and all arithmatic is done with the pointer `offset()` method.

### DIFF
--- a/src/memory_map.rs
+++ b/src/memory_map.rs
@@ -13,7 +13,9 @@ impl MemoryMapTag {
         let start_area = (&self.first_area) as *const MemoryArea;
         MemoryAreaIter {
             current_area: start_area,
-            last_area: ((self_ptr as u32) + self.size - self.entry_size) as *const MemoryArea,
+            last_area: unsafe {
+                self_ptr.offset((self.size - self.entry_size) as isize) as *const MemoryArea
+            },
             entry_size: self.entry_size,
         }
     }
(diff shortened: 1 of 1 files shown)


src/memory_map.rs
```diff
@@ -13,7 +13,9 @@ impl MemoryMapTag {
         let start_area = (&self.first_area) as *const MemoryArea;
         MemoryAreaIter {
             current_area: start_area,
-            last_area: ((self_ptr as u32) + self.size - self.entry_size) as *const MemoryArea,
+            last_area: unsafe {
+                self_ptr.offset((self.size - self.entry_size) as isize) as *const MemoryArea
```
phil-opp: The argument of the [offset](https://doc.rust-lang.org/nightly/std/primitive.pointer.html#method.offset) method is not in units of bytes:

> `unsafe fn offset(self, count: isize) -> *const T`
> Calculates the offset from a pointer. `count` is in units of `T`; e.g. a `count` of 3 represents a pointer offset of `3 * sizeof::<T>()` bytes.

So this change would lead to a wrong pointer.

src/memory_map.rs
```diff
@@ -41,8 +43,9 @@ impl Iterator for MemoryAreaIter {
             None
         } else {
             let area = unsafe{&*self.current_area};
-            self.current_area = ((self.current_area as u32) + self.entry_size)
-                as *const MemoryArea;
+            self.current_area = unsafe {
+                (self.current_area).offset(self.entry_size as isize)
```
phil-opp: Same as above

4e554c4c:
Sorry about that. I didn't test it the first time because my project is in a state of disarray. I tried it with blog_os this time and it seems to work fine.

phil-opp:
No worries :). Thanks!

PR #16, opened 2016-10-12
Make StringTable public

4e554c4c:
I missed this when pushing the commit, StringTable should be
pub so that the desired functionality of "get your own String
Table and change the address" can work.

I know that this changes the public interface, sorry for opening
it so soon after the release.


phil-opp:
Thanks!

> I know that this changes the public interface, sorry for opening
it so soon after the release.

No problem at all! Published as version 0.3.0.

PR #17, opened 2016-10-31
Add `#[derive(Debug)]` on all public memory map structs

cderwin:


phil-opp:
Thanks a lot!

phil-opp:
Published as version 0.3.1

PR #18, opened 2017-02-01
Update README.md

ahmedcharles:


phil-opp:
Nice find!

I don't like the “the reason is because GRUB2 does it” formulation, so I think we should reformulate this paragraph. Maybe something like:

> Note that this format differs […]: The […] fields are `u32` instead of `u16`. The multiboot2.h file […] also specifies these fields as being `u32`, which suggests that the `u16` fields are an editting error. The GRUB2 bootloader [uses `u32` fields](https://github.com/josefbacik/grub2/blob/96695ad84ce9c93f057ba53ae77d04d8561586e9/include/multiboot2.h#L298-L300), too.

ahmedcharles:
Sure, that's fine. Though, the "GRUB2 does it this way" is basically the same, since the header file in the spec is taken directly from the GRUB codebase. :P

phil-opp:
Oh ok. So we have the following situation:

- the specification and the example header file in the specification contradict each other
- the example header file seems to be correct since it's used in GRUB2

ahmedcharles:
Updated the patch.

phil-opp:
Thanks a lot, @ahmedcharles!

PR #19, opened 2017-02-02
Add the command line tag

iBelieve:
This is just a copy of the boot loader name tag code, but for the command line tag (type = 1).


phil-opp:
Thanks!

phil-opp:
Published as version 0.3.2

PR #20, opened 2017-02-13
Implement Debug for BootInformation

ahmedcharles:


4e554c4c:
This looks like quite a lot of information to be printing by default. Is there any reason that you want to implement `fmt::Display` for this instead of `fmt::Debug`?

ahmedcharles:
I'd be fine with implementing Debug instead of Display.

phil-opp:
Yeah, I agree that `Debug` would be better here. Otherwise it looks good to me.

4e554c4c:
Hmm, this also seems to suffer from the "won't work on higher half kernels" problem that #14 had. I'm not quite sure how to fix that though, because `fmt` traits don't take arguments.

ahmedcharles:
Yes, I know that this doesn't work with higher half kernels. The solution in #14 requires unsafe code and that the user of the `BootInformation` (or some subset of it) know what the offset is. I'd prefer to keep knowledge of the kernel offset to be as limited as possible.

The only way to achieve that with this crate is so store the address and offset, then use the offset whenever an absolute address is dereferenced. I'm going to change my local copy of this to do that soon and I'll open a PR to get opinions when it's working well.

ahmedcharles:
Updated it to impl `Debug`.

phil-opp:
>The only way to achieve that with this crate is so store the address and offset, then use the offset whenever an absolute address is dereferenced.

Sounds good to me.

phil-opp:
Thanks a lot, @ahmedcharles!

PR #21, opened 2017-03-08
Add some basic tests.

ahmedcharles:


ahmedcharles:
I updated with a slightly less basic test based on the MBI info from my kernel running in QEMU after being booted by GRUB2.

phil-opp:
Thanks a lot, this looks great!

The grub2 test seems to be failing though.

ahmedcharles:
That's because I don't know if a good way to force the stack to be 8 byte aligned... any ideas?

phil-opp:
Not really… There is https://github.com/rust-lang/rust/issues/33626, but it is not implemented yet. Maybe it's better to leave out the `address & 0b111 == 0` assertion for now.

ahmedcharles:
Hmm... perhaps. I assume x86 is fine with unaligned reads in all the cases where the types would require alignment but for which the alignment would be off.

phil-opp:
> running 5 tests
error: process didn't exit successfully: `/home/travis/build/phil-opp/multiboot2-elf64/target/debug/deps/multiboot2-3552907508445da3` (signal: 11, SIGSEGV: invalid memory reference)

It seems like we're dereferencing an invalid pointer…

ahmedcharles:
The issue was that in the tag iterator, it assumes starting 8 byte alignment and realigns the address rather than realigning the size of the previous tag, which doesn't assume 8 byte alignment overall.

src/header.rs
```diff
@@ -18,8 +18,7 @@ impl Iterator for TagIter {
             tag => {
                 // go to next tag
                 let mut tag_addr = self.current as usize;
-                tag_addr += tag.size as usize;
-                tag_addr = ((tag_addr-1) & !0x7) + 0x8; //align at 8 byte
+                tag_addr += ((tag.size + 7) & !7) as usize; //align at 8 byte
```
phil-opp: So the old code only worked if the `tag_addr` was really 8 byte aligned before? And the new code also works for the misaligned test data?

ahmedcharles: Yes.

phil-opp:
Thanks a lot!

PR #23, opened 2017-03-28
Add support for ELF32 format

jmi2k:
I'm following your guide, but instead I'm making a 32-bit kernel which uses ELF32 files. I added a new feature, `elf32`, which tweaks `ElfSections` to match this standard.

(I know, the project is called multiboot-elf64. But this doesn't hurt and helps a lot :-) )

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,6 @@ description = "An experimental Multiboot 2 crate for ELF-64 kernels."
 [dependencies.bitflags]
 version = "0.4.0"
 features = ["no_std"]
+
+[features]
+elf32 = []
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -149,7 +149,7 @@ pub enum ElfSectionType {
 type ElfSectionFlagsType = u32;
 
 #[cfg(not(feature = "elf32"))]
-type ElfSectionFlagsType = u32;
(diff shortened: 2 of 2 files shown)


4e554c4c:
Does this work with all current methods? `MemoryArea` currently uses 64bit wide fields which I'd assume would not be used by 32bit kernels.

jmi2k:
AFAIK they are independent. The bootloader gives you a 64-bit memory map, no matter the ELF standard you're using.

4e554c4c:
Ah ok that makes sense. I misunderstood the pr.

ahmedcharles:
@jmi2k Would you mind adding a test case based on the multiboot section of your bootloader, similar to the grub2 test case? I suspect that it would be possible to detect elf32 vs elf64 at runtime using the entsize field in the elf symbols tag (it should be 64 bytes for elf64 and 40 for elf32). We currently use the entsize to determine the next entry, but don't verify that the entry is the same size as the struct we use to read it.

jmi2k:
@ahmedcharles Sorry, but I don't know how to do it. Actually, I don't know how to gather all the information needed! I'll try to do it, but don't expect results.

ahmedcharles:
I just printed the memory at the multiboot header address to the screen using the `Debug` impl for `&[u8]` and then just typed the results into the test case.

PR #24, opened 2017-03-29
Change pub fields into functions.

ahmedcharles:
This hides the implementation enough to allow supporting higher memory kernels.

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -4,7 +4,7 @@
 pub struct ElfSectionsTag {
     typ: u32,
     size: u32,
-    pub number_of_sections: u32,
+    number_of_sections: u32,
     entry_size: u32,
     shndx: u32, // string table
     first_section: ElfSection,
@@ -83,10 +83,10 @@ impl Iterator for ElfSectionIter {
 pub struct ElfSection {
     name_index: u32,
     typ: u32,
-    pub flags: u32,
-    pub addr: u32,
+    flags: u32,
+    addr: u32,
     offset: u32,
-    pub size: u32,
+    size: u32,
     link: u32,
     info: u32,
     addralign: u32,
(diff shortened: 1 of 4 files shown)


src/elf_sections.rs
```diff
@@ -82,17 +82,34 @@ impl Iterator for ElfSectionIter {
...
+            3 => ElfSectionType::StringTable,
+            4 => ElfSectionType::RelaRelocation,
+            5 => ElfSectionType::SymbolHashTable,
+            6 => ElfSectionType::DynamicLinkingTable,
+            7 => ElfSectionType::Note,
+            8 => ElfSectionType::Uninitialized,
+            9 => ElfSectionType::RelRelocation,
+            10 => ElfSectionType::Reserved,
+            11 => ElfSectionType::DynamicLoaderSymbolTable,
+            _ => panic!(),
```
4e554c4c: Perhaps it would not be best to panic here because environment and processor specific flags are allowed.

ahmedcharles:
Updated.

4e554c4c:
Hey, since I don't have write access can you update this commit with the most recent master?

ahmedcharles:
I was just doing that and I made a comment on the other commit, since I think it can be implemented at runtime, which would be significantly nicer.

ahmedcharles:
It's updated.

ahmedcharles:
That would prevent ElfSectionType from being `#[repr(u32)]`. I suspect it would be better to just write a custom type, but I'd rather do that in a separate PR. I've added `section_type_raw()` as a workaround, so that information isn't lost in the mean time.

4e554c4c:
Looks good to me 👍

le-jzr:
wrt ElfSectionType, repr(u32) is largely pointless here. We cannot convert from u32 to ElfSectionType, because that's undefined behavior, and the opposite direction we don't need.

PR #25, opened 2017-03-29
Detect ELF-32 at runtime, rather than compile time.

ahmedcharles:
WIP: don't merge.

This is sort of a hacked up runtime implementation of ELF-32 support, primarily so @jmi2k can take a look and see if it works and satisfies their requirements.

I'm planning other changes that will make this much cleaner, but they're not ready yet.

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -1,86 +1,90 @@
+use header::Tag;
 
 #[derive(Debug)]
-#[repr(packed)] // repr(C) would add unwanted padding before first_section
 pub struct ElfSectionsTag {
-    typ: u32,
-    size: u32,
+    inner: *const ElfSectionsTagInner,
+}
+
+pub fn elf_sections_tag(tag: &Tag) -> ElfSectionsTag {
+    assert_eq!(9, tag.typ);
+    let es = ElfSectionsTag {
+        inner: unsafe { (tag as *const _).offset(1) } as *const _,
+    };
+    assert!((es.get().entry_size * es.get().shndx) <= tag.size);
+    es
+}
+
+#[derive(Debug)]
+#[repr(C, packed)] // only repr(C) would add unwanted padding at the end
+struct ElfSectionsTagInner {
     number_of_sections: u32,
     entry_size: u32,
     shndx: u32, // string table
-    first_section: ElfSection,
 }
 
 impl ElfSectionsTag {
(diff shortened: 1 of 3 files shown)


jmi2k:
I made a quick test and it works flawless. Good job!

src/elf_sections.rs
```diff
@@ -53,22 +57,27 @@ impl StringTable {
...
     type Item = &'static ElfSection;
+
     fn next(&mut self) -> Option<&'static ElfSection> {
         if self.remaining_sections == 0 {
             None
         } else {
-            let section = self.current_section;
+            let section: &'static ElfSection = match self.entry_size {
+                40 => unsafe { &*(self.current_section as *const _ as *const ElfSection32) },
+                64 => unsafe { &*(self.current_section as *const _ as *const ElfSection64) },
```
4e554c4c: Is the  `as *const _` nesissary in these lines?

4e554c4c: I would prefer `core::mem::size_of<ElfSection32|64()` to comparing with `40` and `60` if they are equivilant.

ahmedcharles: I have more changes I'd like to get in that would clean this one up more, but yes, you have to cast twice, once to get a raw pointer of the same type and again to change the type.

ahmedcharles: That isn't possible because `match` doesn't accept `size_of` in a pattern.

ahmedcharles:
I updated this to be closer to reasonable. Input welcome.

robert-w-gries:
This PR has been great in my kernel that supports both i686 and x86_64 targets.  Any update on whether this will be merged?

phil-opp:
I'm sorry, I didn't have the time for maintaining this crate over the past months. I'm happy to add maintainers to this project, if someone is willing, to avoid such delays in the future.

phil-opp:
Thanks a lot for this PR, @ahmedcharles! From a quick look, it looks good. I don't use 32bit ELFs personally, so I don't have a good way of testing this feature. I don't have the time to do an in-detail review either. Nevertheless, I'm going to merge this PR since it seems to work, according to the comments in this thread.

I also decided to give you push access to this repo. Feel free to push and merge any non breaking changes at your own discretion.

ahmedcharles:
Thanks.

PR #29, opened 2017-05-11
Make MemoryMapIter be Send

tedsta:
Hey thanks for the great OS series!

MemoryMapIter is used in the AreaFrameAllocator, which I want to put inside a Mutex and make global. I want to access it in syscalls. To be put inside a Mutex, it has to be `Send`.


phil-opp:
Sorry for the delay. Looks good to me!

Issue #22, opened 2017-03-27
ModuleTag::name() and others are unsafe


zarevucky:
The comment in the function tries to explain safety with "The multiboot specification defines the module str as valid utf-8, therefore this function produces defined behavior". Such trust in constraint assumed by specification is a horribly wrong thing to do.

It might not matter much when the data is used by kernel to boot, as corrupted data there automatically break all safety, but that makes the implicit assumption that the crate will never, ever, ever, be used to access multiboot data in other contexts, e.g. a memory dump for debugging.

zarevucky:
Well, I guess the entire implementation has that problem. Mind if I start another crate that tries to be a bit more generic? :)

4e554c4c:
Good point. We rely on the multiboot specification to be correct in multiple contexts and it would probably be best to assume that data can be malformed. The most dangerous assumptions, in my opinion, are not assuming utf8 is correct, but trusting `size` sections, which could easily lead to segfaults or worse in a kernel context.

zarevucky:
To elaborate on my previous comment, Multiboot is particular in that the structures are linked together with physical addresses, and the crate basically interprets them all as raw pointers. Alternative is to abstract away the memory as a random-access file and let user decide how those reads are fulfilled. That imo could make the crate cleaner and less riddled with unsafe blocks.

4e554c4c:
@zarevucky That sounds like a good idea! Treating data in this way has created problems before as well. It may make sense to create your own crate because the small scope of this one, however if you have ideas on how to achieve these goals feel free to impliment them here.

phil-opp:
> Well, I guess the entire implementation has that assumption. Mind if I start another crate that tries to be a bit more generic? :)

Not at all! I don't have the time to maintain this crate anyway, so I would be happy to switch to something better. Of course, we could also reuse this crate, since semantic versioning allows us to do breaking changes. I just gave both of you push access, feel free to do any refactoring and restructuring you want!

ahmedcharles:
[...]

PR #31, opened 2017-08-27
Make TagIter use explicit lifetimes

Columbus240:
This is a move away from the many 'static lifetimes, as proposed in issue #22.

Small question regarding the specific implementation:
I made the `phantom` field of `TagIter` public, because a `TagIter` is constructed in `src/lib.rs` and I didn’t want to introduce a `TagIter::new()` method without approval. I don’t know which is better style here: making `phantom` public, as done in this PR, or make a `new()` function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 #![no_std]
 
 use core::fmt;
+use core::marker::PhantomData;
 
 use header::{Tag, TagIter};
 pub use boot_loader_name::BootLoaderNameTag;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -80,12 +81,12 @@ impl BootInformation {
...
     }
 
-    fn get_tag(&self, typ: u32) -> Option<&'static Tag> {
+    fn get_tag<'a>(&'a self, typ: u32) -> Option<&'a Tag> {
         self.tags().find(|tag| tag.typ == typ)
     }
 
     fn tags(&self) -> TagIter {
-        TagIter { current: &self.first_tag as *const _ }
+        TagIter { current: &self.first_tag as *const _, phantom: PhantomData }
```
Columbus240: Because of this line I made `phantom` public. This is the only place where an `fn new(tag: *const Tag) -> TagIter` would be necessary.

xacrimon:
Been a while. Is this pull request still useful?

Columbus240:
(Note that I mostly forgot the specifics)
I think the point about using less `'static` still stands. Issue #22 doesn’t seem to have advanced.
To start a little bikeshedding, I think (now), that an `fn new(…)` would be appropriate.
I haven’t yet looked over the current state of the codebase and the changes in the last 2 years.
If you’d like, I can take the time and adapt the patch to resolve the conflicts.

xacrimon:
That would be hugely helpful @Columbus240

Columbus240:
Closing this PR in favor of a new one, because in the mean time, I deleted my fork on GitHub and can’t access this branch anymore.

PR #33, opened 2017-12-05
Update to the 1.0.0 version of bitflags.

ahmedcharles:
Note, this is technically a breaking change but I'd say it's a good one given that it results in using a stable version of bitflags.


4e554c4c:
Lgtm, since it's breaking you should increase the version though.

ahmedcharles:
There were already some breaking changes (changing fields into functions) and I want to propose a few more which should clean things up. So, I'll just make a commit at the end that changes the version before releasing.

4e554c4c:
Thanks!

PR #35, opened 2017-12-06
Fix elf sections so it returns the last section.

ahmedcharles:


robert-w-gries:
Can you add a testcase that explicitly tests the number of elf sections?  Maybe re-use the input data from the grub test, so we don't blow up the size of src/lib.rs.

ahmedcharles:
Added a new test focused on elf sections. It only has one section, so the bug is painfully obvious.

xacrimon:
Have you tested? Short on time and can't test myself.

xacrimon:
@ahmedcharles

ahmedcharles:
Yeah, I wouldn't add a test without running it. Also, CI passed.

ahmedcharles:
I've also tested it in a kernel.

robert-w-gries:
@xacrimon Do you have permissions for merging?  If so, you can probably go ahead and merge the reviewed PRs

ahmedcharles:
Thanks.

PR #36, opened 2017-12-06
Various cleanups.

ahmedcharles:
Add debug implementations.
Mark everything as repr(C) and repr(packed), just to be sure.

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -45,7 +45,7 @@ impl ElfSectionsTag {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct ElfSectionIter {
     current_section: *const u8,
     remaining_sections: u32,
@@ -78,6 +78,7 @@ impl Iterator for ElfSectionIter {
     }
 }
 
+#[derive(Debug)]
 pub struct ElfSection {
     inner: *const u8,
     string_section: *const u8,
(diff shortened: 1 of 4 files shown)


robert-w-gries:
👍 LGTM

----------------

On an semi-related note, I'd appreciate help in understanding what's going on with `repr(packed)`.

This note from the [`repr` documentation](https://doc.rust-lang.org/beta/nomicon/other-reprs.html#reprpacked) is intimidating:

>In particular, most architectures strongly prefer values to be aligned. This may mean the unaligned loads are penalized (x86), or even fault (some ARM chips). For simple cases like directly loading or storing a packed field, the compiler might be able to paper over alignment issues with shifts and masks. However if you take a reference to a packed field, it's unlikely that the compiler will be able to emit code to avoid an unaligned load.

> As of Rust 1.0 this can cause undefined behavior.
> repr(packed) is not to be used lightly. Unless you have extreme requirements, this should not be used.

Since our case requires liberal use of `repr(C, packed)`, is there a way to minimize penalties and undefined behavior?  Perhaps through more asserts or wrappers?

Let me know if there's something to this, and I'll open a separate issue.

phil-opp:
As I understand it, repr(C) should suffice as long as the fields are aligned. With repr(packed) everything should still work without undefined behavior as long as we avoid references to unaligned fields.

https://github.com/rust-lang/rust/pull/44884 was merged about a week ago, so this should make things safer (but I don't know details either).

ahmedcharles:
I wish Rust had static assert, which would allow asserting the size of a struct to ensure there is no padding. I'll probably look into that and switch to using that as a way to verify that what you see is what you get, since using unsafe to access fields is somewhat silly in many cases.

phil-opp:
We could add a unit test that checks the struct sizes.

PR #37, opened 2018-01-04
Resolve tyvar_behind_raw_pointer warning

robert-w-gries:
Fix the following warning before it becomes a hard error:

```
warning: the type of this value must be known in this context
  --> src/elf_sections.rs:11:43
   |
11 |         inner: unsafe { (tag as *const _).offset(1) } as *const _,
   |                                           ^^^^^^
   |
   = note: #[warn(tyvar_behind_raw_pointer)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #46906 <https://github.com/rust-lang/rust/issues/46906>
```


phil-opp:
Thanks a lot!

PR #46, opened 2018-06-03
Add Framebuffer Tag

tuomas56:
Adds the `BootInformation::framebuffer_tag` function along with associated structs, for framebuffer information as well as color data.

### DIFF
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -2,7 +2,7 @@ use core::ptr;
 use core::slice;
 use header::Tag;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct FramebufferTag<'a> {
     pub address: u64,
     pub pitch: u32,
@@ -12,7 +12,7 @@ pub struct FramebufferTag<'a> {
     pub buffer_type: FramebufferType<'a>
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum FramebufferType<'a> {
     Indexed {
         palette: &'a [FramebufferColor]   
@@ -25,13 +25,13 @@ pub enum FramebufferType<'a> {
     Text
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct FramebufferField {
     position: u8,
     size: u8
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 #[repr(C, packed)]
 pub struct FramebufferColor {
     red: u8,
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -1,0 +1,124 @@
+use core::ptr;
+use core::slice;
+use header::Tag;
+
+#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


phil-opp:
Thanks for the PR @tuomas56!

@rust-osdev/multiboot2 Has anyone time to do a review?

src/framebuffer.rs
```diff
@@ -0,0 +1,123 @@
+use core::ptr;
+use core::slice;
+use header::Tag;
+
+#[derive(Debug, PartialEq)]
+pub struct FramebufferTag<'a> {
```
robert-w-gries: Shouldn't this be `repr(C, packed)`?

src/framebuffer.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Debug, PartialEq)]
+#[repr(C, packed)]
+pub struct FramebufferColor {
+    pub red: u8,
+    pub green: u8,
+    pub blue: u8
+}
+
+struct Reader {
```
robert-w-gries: This seems like a useful utility for other tags.  Can you move it into `src/lib.rs` and make it `pub`?

src/framebuffer.rs
IsaacWoods: As far as I can see, this doesn't follow the representation in memory and is created using `Reader`, so it doesn't need to be

src/framebuffer.rs
```diff
@@ -0,0 +1,123 @@
...
+    let buffer_type = match type_no {
+        0 =>  {
+            let num_colors = reader.read_u32();
+            let palette = unsafe {
+                slice::from_raw_parts(reader.ptr.offset(reader.off as isize) as *const FramebufferColor, num_colors as usize)
+            } as &'static [FramebufferColor];
+            FramebufferType::Indexed { palette }
+        },
+        1 => {
+            let red_pos = reader.read_u8();     //These refer to the bit positions of the MSB of each field
```
IsaacWoods: Very pedantic nit: these should have spaces at the beginning of the comment

src/framebuffer.rs
IsaacWoods: Looks super useful. I wonder if it should be `pub(crate)` tho, as it doesn't seem useful to the user of the library

robert-w-gries: Yes, `pub(crate)` seems like the right choice here

src/framebuffer.rs
robert-w-gries: Yeah you're right.  `Reader` is nice in that regard.

The only place we need `repr(C, packed)` is `FrameBufferColor` since we grab a slice of that memory, and it's already there.

Sorry for the noise!

robert-w-gries:
@IsaacWoods Can you take a look at the changes?

IsaacWoods:
Looks good to me; ready for merge as far as I’m concerned. Thanks @tuomas56!

phil-opp:
@rust-osdev/multiboot2 Maybe release a new crates.io version? You should all have publish rights.

PR #50, opened 2018-06-27
Return memory_map entries as u64, to support 32-bit targets

### LINKED ISSUE #49: i386 kernels cannot use MemoryMapTag safely
I'm currently writing an i386 kernel, and am using `multiboot2` to parse the multiboot2 information. Unfortunately, the MemoryMapTag returns the `start_address`/`end_address` as a usize, whereas they are stored as u64. https://github.com/rust-osdev/multiboot2-elf64/blob/master/src/memory_map.rs#L33.

For 32-bit kernels, this means that if a PC has more than 4GB of available RAM, we'll get some completely wrong values, and end up thinking that a range from `0x000000-0xXXXXXX` is available, wrecking havoc. Those should be sent to return u64s (which I did locally in my kernel), but that would be a breaking change.

roblabla:
Fixes #49 

Note: This is a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "multiboot2"
-version = "0.6.0"
+version = "0.7.0"
 authors = ["Philipp Oppermann <dev@phil-opp.com>", "Calvin Lee <cyrus296@gmail.com>"]
 license = "MIT/Apache-2.0"
 description = "An experimental Multiboot 2 crate for ELF-64/32 kernels."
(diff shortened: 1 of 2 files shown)


roblabla:
I think we should bump to 0.7.0 ? My understanding (might be wrong) is that specifying `0.6.0` in Cargo.toml will make cargo auto-update crates to `0.6.x` (Cargo.lock notwithstanding). So bumping to `0.6.1` runs the risk of breaking people's code the next time they run cargo build. 

To avoid this, we might want to bump to 0.7 ?

IsaacWoods:
I think it'd take a `cargo update` to break it, but that's probably a better idea. Can you bump it to `0.7.0` then?

IsaacWoods:
Great, thanks @roblabla!

PR #51, opened 2018-06-27
Allow users to read elf section alignment

Ben-Alderson:
Adds an addralign() function to `ElfSection`, which returns the alignment of that section.

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -169,6 +169,10 @@ impl ElfSection {
         self.get().size()
     }
 
+    pub fn addralign(&self) -> u64 {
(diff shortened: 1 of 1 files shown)


IsaacWoods:
I think I’ve already published #52 as `0.7.1`, so we’ll need to bump the version again to publish this

phil-opp:
Ah ok, I was just confused because the GitHub commit overview is based on commit times and not commit order.

phil-opp:
> so we’ll need to bump the version again to publish this

There are no other incoming changes at the moment, so I think we should do this to make this change available to crates.io users.

PR #52, opened 2018-06-28
Return u64 from MemoryArea::size as well

IsaacWoods:
Looks like I was a bit hasty merging #50 and missed that we're still returning a `usize` from the `size` method. I've also bumped the version to `0.7.1` from `0.7.0`, as this only patches the previous breaking change.

@rust-osdev/multiboot2 could I get a review so I can publish the new version (`0.7.0` is already published, it'd be good to minimise the number of people who upgrade and have to work around this)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "multiboot2"
-version = "0.7.0"
+version = "0.7.1"
 authors = ["Philipp Oppermann <dev@phil-opp.com>", "Calvin Lee <cyrus296@gmail.com>"]
 license = "MIT/Apache-2.0"
 description = "An experimental Multiboot 2 crate for ELF-64/32 kernels."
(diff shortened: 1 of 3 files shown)


robert-w-gries:
Do we also need to remove all uses of `usize` from the library?

[See here](https://github.com/rust-osdev/multiboot2-elf64/search?q=usize&unscoped_q=usize)

IsaacWoods:
So, looking into each case, I *think* most of them should be okay, because afaik, virtual addresses on x86 are always 32-bit. The problem is physical addresses when PAE is enabled, because then physical addresses can be up to 64-bits (but are actually 36-bits).

As far as I can tell, the only other case we're working with physical addresses is the RSDP code. We're actually doing it correctly, because if a 64-bit XSDT address is specified on a 32-bit arch, we should truncate it to 32-bits. That could use a comment though.

PR #54, opened 2019-02-23
Add support for multiboot information at an offset.

ahmedcharles:


xacrimon:
Can't merge. Please sync the base branch to yours.

ahmedcharles:
Should be the same as before with some adjustments to work with the other PR.

xacrimon:
Merged

ahmedcharles:
Thanks.

phip1611:
For API simplifications and streamlining, I'm planning to deprecate the "with offset" API. I could not find a valid reason for this "more complicated" API, as it is a no-brainer to perform a `ptr.add(offset)` as a developer and just use the regular load function.

https://github.com/rust-osdev/multiboot2/pull/142

I'm not going to merge this today or tomorrow, so that people have
- chances to explain to me if I miss something and this API is indeed very important
- I have more time to see if I missed something

PR #57, opened 2019-04-30
Use no `'static` lifetimes anymore

Columbus240:
Redid PR #31.
The code builds and passes all automated tests of the repo, but I didn’t test it using another project or similar. The many unsafe blocks doing pointer dereferencing and such hide stuff from the lifetime-checker. So simply because it builds is not yet assuring to me. Please check and test before merging.

I made an API change to `TagIter`, requiring the use of a constructor function.

I am sure that I haven’t captured the intention of issue #22 completely. Although, all handed-out references have some parametric lifetime, this patch doesn’t improve “safety”. Meaning: out-of-bounds accesses aren’t prevented and the multiboot2 header can’t be at an arbitrary place in memory, because the physical addresses inside the header are still directly interpreted as pointers.

### DIFF
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -55,7 +55,7 @@ pub fn framebuffer_tag<'a>(tag: &'a Tag) -> FramebufferTag<'a> {
             let num_colors = reader.read_u32();
             let palette = unsafe {
                 slice::from_raw_parts(reader.read_u32() as usize as *const FramebufferColor, num_colors as usize)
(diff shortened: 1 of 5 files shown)


xacrimon:
Please update this branch

xacrimon:
@Columbus240

IsaacWoods:
LGTM, thanks @Columbus240!

I'll merge this, and publish as version 0.8.0 (I think I have permissions to do that)

PR #59, opened 2019-05-12
Publish v0.8.0

IsaacWoods:
This includes the changes from #57.

@rust-osdev/multiboot2 could I get a review here, after which I'll publish the changes to crates.io


IsaacWoods:
Thanks @robert-w-gries 

` 269b5a6` published as v0.8.0

Issue #63, opened 2020-06-12
RSDP checksum and ext_checksum are not particularly useful


dlrobertson:
# Summary

A correct calculation of the RSDP checksum requires access to enum members that are private or reserved.

# Description

AFAIK the [RsdpV1Tag::checksum](https://github.com/rust-osdev/multiboot2-elf64/blob/master/src/rsdp.rs#L35), [RsdpV2Tag::checksum](https://github.com/rust-osdev/multiboot2-elf64/blob/master/src/rsdp.rs#L83), and [RsdpV2Tag::ext_checksum](https://github.com/rust-osdev/multiboot2-elf64/blob/master/src/rsdp.rs#L105) are not particularly useful in their current form. From ACPI 6.3 Section 5.2.5.3 Table 5-27:

The Checksum

 > This is the checksum of the fields... This includes only the first 20 bytes of this table, bytes 0 to 19, including the checksum field. These bytes must sum to zero.

The Extended Checksum

> This is a checksum of the entire table, including both checksum fields.

A correct calculation of either checksum requires access to private members of the `RsdpV1Tag` and `RsdpV2Tag` for which there is no access function defined.

# Possible Solution

Instead of surfacing these members like `RsdpV2Tag::_rsdt_address` and `RsdpV2Tag::length` provide the `RsdpV1Tag::verify_checksum`, `RsdpV2Tag::verify_checksum`, and `RsdpV2Tag::verify_ext_checksum` functions that perform the calculation.

dlrobertson:
I have a simple implementation working locally. If the proposed approach seems reasonable, I can post a PR.

phil-opp:
@rust-osdev/multiboot2 Does anyone have time to look into this issue?

IsaacWoods:
Hi - I seem to have written this code so am happy to look into this.

Yes the current getters seem pretty useless and replacing them with some sort of validation method seems sensible to me.

However, for revision 2.0 and above (the V2 tag), there's no need to validate both checksums, instead you just add the entire contents of the table and see if it sums to `0`. For an example of how `acpi` validates both versions of the RSDP, see [this method](https://github.com/rust-osdev/acpi/blob/master/acpi/src/rsdp.rs#L40-L78). It seems sensible to just have a `validate` method on both tags that does the right thing.

PR #64, opened 2020-06-28
rsdp: add the validate_checksum to rsdp

dlrobertson:
Add the validate_checksum method to the implementation of the RsdpV1Tag
and RsdpV2Tag.

CC: @phil-opp @IsaacWoods 
Fixes: #63

### DIFF
--- a/src/rsdp.rs
+++ b/src/rsdp.rs
@@ -6,7 +6,10 @@
 /// Even though the bootloader should give the address of the real RSDP/XSDT, the checksum and
 /// signature should be manually verified.
 
+use core::slice;
 use core::str;
+
+const RSDPV1_LENGTH: usize = 20;
 
 /// This tag contains a copy of RSDP as defined per ACPI 1.0 specification. 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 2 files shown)


src/rsdp.rs
```diff
@@ -77,11 +85,16 @@ impl RsdpV2Tag {
         str::from_utf8(&self.signature).ok()
     }
 
-    /// The value to add to all the other bytes (of the Version 1.0 table) to calculate the Checksum of the table.
-    ///
-    /// If this value added to all the others and casted to byte isn't equal to 0, the table must be ignored. 
-    pub fn checksum(&self) -> u8 {
-        self.checksum
+    /// Validation of the RSDPv2 extended checksum
+    pub fn checksum_is_valid(&self) -> bool {
```
dlrobertson, line 89: Only added a validation of the RSDPv2 _extended_ checksum for the RsdpV2Tag. Should we have a RSDPv1 checksum validation and extended checksum validation function for the RsdpV2Tag?

src/rsdp.rs
```diff
@@ -29,11 +32,16 @@ impl RsdpV1Tag {
...
+    /// Validation of the RSDPv1 checksum
+    pub fn checksum_is_valid(&self) -> bool {
+        let result = unsafe {
+            let bytes = slice::from_raw_parts(
+                self as *const _ as *const u8,
+                RSDPV1_LENGTH + 8
+            );
+            bytes[8..].iter().fold(0usize, |acc, val| acc.wrapping_add(*val as usize))
+        };
+        (result & 0xff) == 0
```
IsaacWoods, line 44: Do you have a reference for why you're summing into a `usize` and then masking for the lowest byte later? My interpretation of the ACPI spec is that you should sum within a single byte:
```
This is the checksum of the fields defined in the ACPI 1.0 specification. This includes only the first 20 bytes of this table, bytes 0 to 19, including the checksum field. These bytes must sum to zero.
```

This would apply to both `checksum_is_valid` methods:
```suggestion
            bytes[8..].iter().fold(0u8, |acc, val| acc.wrapping_add(*val))
        };
        result == 0
```

dlrobertson: yeah all we need is a u8 and that would remove the need for the mask later. I think in the original version i didn't use wrapping add, so the use of usize was to avoid an overflow.

src/rsdp.rs
dlrobertson: Yeah, the checksum of the ACPI 1.0 fields is still valid in the a RSDPv2 structure, but I don't know of a reason for a user to only check this. If we do find a reason to include it, it shouldn't be hard to add.

src/rsdp.rs
dlrobertson: Updated

IsaacWoods:
Published as multiboot2 v0.9.0

PR #68, opened 2020-08-24
Improve usage of repr(packed)

### LINKED ISSUE #38: Warning safe_packed_borrows will become a hard error
### Warning

While restarting a garbage Travis build, I noticed the following warnings:

```
warning: #[derive] can't be used on a non-Copy #[repr(packed)] struct (error E0133)
 --> src/header.rs:1:10
  |
1 | #[derive(Debug)]
  |          ^^^^^
  |
  = note: #[warn(safe_packed_borrows)] on by default
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #46043 <https://github.com/rust-lang/rust/issues/46043>

warning: borrow of packed field requires unsafe function or block (error E0133)
 --> src/elf_sections.rs:9:5
  |
9 |     assert_eq!(9, tag.typ);
  |     ^^^^^^^^^^^^^^^^^^^^^^^
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #46043 <https://github.com/rust-lang/rust/issues/46043>
  = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

There are around 20 more `#[derive(Debug)]` warnings that I snipped for brevity.

### `#[repr(packed)]`

The warning is related to our usage of `#[repr(packed)]` in the crate.  There is undefined behavior associated with derefencing unaligned structs, which rust-lang/rust#46043 will prevent with a compiler error.

### How to Fix

From the rust issue:

>The most common ways to fix this warnings are:
> 1. remove the #[repr(packed)] attribute if it is not actually needed. A few crates had unnecessary #[repr(packed)] annotations - for example, tendril.
> 2. copy the fields to a local first. When accessing a field of a packed struct directly without using a borrow, the compiler will make sure the access is done correctly even when the field is unaligned. The then-aligned local can then be freely used. For example:
```rust
let x = Foo { start: 0, data: Unaligned(1) };
let temp = x.data.0;
println!("{}", temp); // works
// or, to the same effect, using an `{x}` block:
println!("{}", {x.data.0}); // works
use(y);
```

> One annoying case where this problem can appear is if a packed struct has builtin derives
>
> If your struct already derives Copy and has no generics, the compiler will generate code that copies the fields to locals and will therefore work. Otherwise, you'll have to write the derive implementations manually.

### Feedback

Before submitting a PR, I'd like feedback on the following:

1. Do our impacted structs **need** to use`#[repr(packed)]`?  It seems that `#[repr(packed)]` should be a last resort instead of a feature liberally used.
1. Can we implement `Copy` for our affected structs? Or do we need to manually implement `Debug` for those structs?

Caduser2020:
Close #38.
We occasionally use`repr(packed)` when the struct is unaligned. When we use it, I added a comment explaining why. In addition, we derive `Copy` and `Clone` to allow the compiler to copy the fields to locals.

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -13,7 +13,7 @@ pub unsafe fn elf_sections_tag(tag: &Tag, offset: usize) -> ElfSectionsTag {
     assert_eq!(9, tag.typ);
     let es = ElfSectionsTag {
         inner: (tag as *const Tag).offset(1) as *const ElfSectionsTagInner,
-        offset: offset,
+        offset,
     };
     assert!((es.get().entry_size * es.get().shndx) <= tag.size);
     es
--- a/src/framebuffer.rs
+++ b/src/framebuffer.rs
@@ -66,7 +66,7 @@ pub struct FramebufferField {
 
 /// A framebuffer color descriptor in the palette.
 #[derive(Clone, Copy, Debug, PartialEq)]
-#[repr(C, packed)]
+#[repr(C, packed)] // only repr(C) would add unwanted padding at the end
 pub struct FramebufferColor {
     /// The Red component of the color.
     pub red: u8,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,7 +115,7 @@ pub struct BootInformation {
 }
 
 #[derive(Clone, Copy)]
-#[repr(C, packed)]
+#[repr(C)]
 struct BootInformationInner {
     total_size: u32,
     _reserved: u32,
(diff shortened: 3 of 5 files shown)


VictorKoenders:
Maybe outside of the scope of this PR, but should there be compile-time validations or tests to make sure the structs are the right size?

In my project i have a couple of
```rust
fn _test_size() {
    unsafe {
        core::mem::transmute::<[u8; 56], Hpet>([0u8; 56]);
    }
}
```

Which give a compile-time check to validate the `Hpet` struct is still 56 bytes.

Caduser2020:
@VictorKoenders I completely agree. I will open a new PR for this later.

PR #71, opened 2020-11-01
Access to non-available memory areas 

CalebLBaker:
https://github.com/rust-osdev/multiboot2-elf64/issues/70

### DIFF
--- a/src/memory_map.rs
+++ b/src/memory_map.rs
@@ -22,16 +22,15 @@ pub struct MemoryMapTag {
 
 impl MemoryMapTag {
     /// Return an iterator over all AVAILABLE marked memory areas.
-    pub fn memory_areas(&self) -> MemoryAreaIter {
-        MemoryAreaIter {
-            iter: self.all_memory_areas(),
-        }
+    pub fn memory_areas(&self) -> impl Iterator<Item = &MemoryArea> {
+        self.all_memory_areas().filter(|entry| entry.typ == 1)
     }
+
     /// Return an iterator over all marked memory areas.
-    pub fn all_memory_areas(&self) -> AllMemoryAreaIter {
+    pub fn all_memory_areas(&self) -> MemoryAreaIter {
         let self_ptr = self as *const MemoryMapTag;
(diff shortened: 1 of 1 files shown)


src/memory_map.rs
```diff
@@ -23,9 +23,15 @@ pub struct MemoryMapTag {
 impl MemoryMapTag {
     /// Return an iterator over all AVAILABLE marked memory areas.
     pub fn memory_areas(&self) -> MemoryAreaIter {
```
IsaacWoods, line 25: I wonder if, rather than having two iteratory types, this could return `impl Iterator<Item = &MemoryArea` (maybe you'll need a lifetime idk) and then the body could be `self.all_memory_areas().filter(|entry| entry.typ == 1)`? I think that'd be cleaner, if it works.

src/memory_map.rs
```diff
@@ -91,26 +97,42 @@ pub enum MemoryAreaType {
     Defective,
 }
 
-/// An iterator over Available memory areas.
+/// An iterator over all memory areas
 #[derive(Clone, Debug)]
-pub struct MemoryAreaIter<'a> {
+pub struct AllMemoryAreaIter<'a> {
```
Caduser2020, line 102: `EFIMemoryMap` also could use an iterator over all memory areas.

src/memory_map.rs
CalebLBaker: That would be cleaner and it does work. Initially I was worried that filter wouldn't work without the standard library (I'm using this crate for work on an OS kernel where I don't have the standard library and I'm new to working without the standard library, so I haven't yet figured out what is part of the standard library vs what is part of core), so I chose to avoid. On further investigation, it seems to work fine, so I will update my pull request to make use of fileter.

src/memory_map.rs
CalebLBaker: It looks to me like EFIMemoryMapTag already iterates over all memory areas.

IsaacWoods (approved):
Great, thank you!

PR #72, opened 2021-03-18
chore: use `impl Iterator`

toku-sa-n:
Two return types are mixed so all return types are integrated to `impl
Iterator`.


IsaacWoods:
Thanks for working on this library! I don't see any inherent problem with doing this, and I agree it makes the API slightly more consistent.

IsaacWoods (approved):

PR #73, opened 2021-03-19
fix(docs): remove a code example in the top page of the doc

toku-sa-n:
The previous version of the `asm!` macro is renamed to `llvm_asm!`. It will not be stable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "multiboot2"
-version = "0.10.0"
+version = "0.10.1"
 authors = ["Philipp Oppermann <dev@phil-opp.com>", "Calvin Lee <cyrus296@gmail.com>"]
 license = "MIT/Apache-2.0"
 description = "An experimental Multiboot 2 crate for ELF-64/32 kernels."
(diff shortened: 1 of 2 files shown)


phil-opp:
I'm not really involved in the maintenance of this crate anymore, but this code snippet seems quite fragile to me. I think we should make it more clear that this example will only work when used at the very beginning of a naked function, otherwise the `ebx` register will likely contain something else already.

IsaacWoods:
I agree with Phil. I think it'd probably be better to get rid of this asm snippet entirely as a footgun, as it's probably something users need to think about themselves to get right (I imagine most will be using a 32-bit stub at the beginning of their kernels anyway, to switch to long mode before entering, and so they need to work out how to pass this pointer through themselves).

@toku-sa-n thanks for working on this! I'll merge this pull-request if it removes the snippet entirely, unless @phil-opp thinks that's a bad approach?

toku-sa-n:
I'll remove it if he accepts.

phil-opp:
I agree that removing the snippet is probably the best solution.

toku-sa-n:
Removed.

IsaacWoods (approved):

PR #74, opened 2021-03-21
style: apply `rustfmt`

toku-sa-n:
I tried to send another PR but thought it would be better to apply `rustfmt` before that not to make a big diff.


IsaacWoods (approved):

IsaacWoods:
Thanks!

PR #75, opened 2021-06-22
make `get_tag` and `tags` public

### LINKED ISSUE #69: Other tags?
I see that both `get_tag` and `tags` are not public, is there any reason for this? I'd love to be able to query what other tags may be in the header.

int0xh:
closes #69

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -227,11 +227,13 @@ impl BootInformation {
         unsafe { &*self.inner }
     }
 
-    fn get_tag<'a>(&'a self, typ: u32) -> Option<&'a Tag> {
+    /// Search for a tag by `type` identifier.
+    pub fn get_tag<'a>(&'a self, typ: u32) -> Option<&'a Tag> {
         self.tags().find(|tag| tag.typ == typ)
     }
 
-    fn tags(&self) -> TagIter {
+    /// Get a iterator over tags passed by bootloader.
+    pub fn tags(&self) -> TagIter {
         TagIter::new(unsafe { self.inner.offset(1) } as *const _)
     }
 }


strcoy:
@IsaacWoods pinging for review! :smile:

IsaacWoods:
Hi @strcoy

Out of interest, is the crate lacking support for a tag you need, or is this just for maximal permissiveness?

phip1611:
@IsaacWoods Should we add this to the v1.0.0 release?  This would fit because we also have to make the tag type enum public.

src/lib.rs
```diff
@@ -227,11 +227,13 @@ impl BootInformation {
         unsafe { &*self.inner }
     }
 
-    fn get_tag<'a>(&'a self, typ: u32) -> Option<&'a Tag> {
+    /// Search for a tag by `type` identifier.
+    pub fn get_tag<'a>(&'a self, typ: u32) -> Option<&'a Tag> {
```
IsaacWoods, line 231: We don't use `get_*` for anything else - I think we should use this as an opportunity to just rename this to `tag`.

phip1611:
As we have separate getters for every single possible tag, which returns Options, I don't see why this should be public. This would be more confusing than benefitical. I'd propose that we check if we have indeed a getter for every single tag type and if not, we do this instead in this PR. Opinion, @IsaacWoods ?

IsaacWoods:
I'm inclined to agree - if a tag type is missing, the preferred answer imo is to add it to this library. With a format like `stivale`, where I think custom tag types are allowed, this would make more sense, but yeah I'm inclined to close unless the original author comes back to us.

int0xh:
> I'm inclined to agree - if a tag type is missing, the preferred answer imo is to add it to this library. With a format like `stivale`, where I think custom tag types are allowed, this would make more sense, but yeah I'm inclined to close unless the original author comes back to us.

Makes sense! Closing it.

PR #76, opened 2021-07-02
much improved debug output of BootInformation + enum TagType

phip1611:
The current debug output of the BootInformation-struct is not optimal. It isn't nicely formatted (no spaces), doesn't include all relevant fields (like UEFI system table tag) and it outputs all ELF-sections, which can be pretty fast thousands of lines with pretty much useless output. 

What do you think about my improvement?

POC with my multiboot2 Rust kernel in QEMU:

image

```
Multiboot2 Boot Information {
    start_address: 0x0000000000597000,
    end_address: 0x00000000005e9ae0,
    total_size: 0x0000000000052ae0,
    boot_loader_name_tag: "GRUB 2.04-1ubuntu44.2",
    command_line: "",
    memory_areas: None,
    module_tags (count): 0,
    elf_sections_tags (count): 5287,
    efi_32_ih: None,
    efi_64_ih: Some(
        EFIImageHandle64 {
            typ: 20,
            size: 16,
            pointer: 114517528,
        },
    ),
    efi_sdt_32_tag: None,
    efi_sdt_64_tag: Some(
        EFISdt64 {
            typ: 12,
            size: 16,
            pointer: 127852568,
        },
    ),
    efi_memory_map_tag: None,
}
```

### DIFF
--- a/src/elf_sections.rs
+++ b/src/elf_sections.rs
@@ -1,4 +1,5 @@
 use header::Tag;
+use core::fmt::{Formatter, Debug};
 
 /// This tag contains section header table from an ELF kernel.
 ///
@@ -41,7 +42,7 @@ impl ElfSectionsTag {
     ///     }
     /// }
     /// ```
-    pub fn sections(&self) -> impl Iterator<Item = ElfSection> {
+    pub fn sections(&self) -> ElfSectionIter {
         let string_section_offset = (self.get().shndx * self.get().entry_size) as isize;
         let string_section_ptr =
             unsafe { self.first_section().offset(string_section_offset) as *const _ };
@@ -64,7 +65,7 @@ impl ElfSectionsTag {
 }
 
 /// An iterator over some ELF sections.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct ElfSectionIter {
     current_section: *const u8,
     remaining_sections: u32,
@@ -93,6 +94,28 @@ impl Iterator for ElfSectionIter {
             }
         }
         None
+    }
+}
+
+impl Debug for ElfSectionIter {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        let mut debug = f.debug_list();
+        self.clone().for_each(|ref e| {
+            debug.entry(e);
+        });
+        debug.finish()
+    }
+}
+
+impl Default for ElfSectionIter {
+    fn default() -> Self {
+        Self {
+            current_section: core::ptr::null(),
+            remaining_sections: 0,
+            entry_size: 0,
+            string_section: core::ptr::null(),
+            offset: 0
+        }
     }
 }
 
--- a/src/header.rs
+++ b/src/header.rs
@@ -1,11 +1,93 @@
 use core::marker::PhantomData;
+use core::fmt::{Debug, Formatter};
+use core::cmp::Ordering;
 
-#[derive(Clone, Copy, Debug)]
+/// Possible Types of a [`Tag`]. The names and values are taken from the example C code
+/// at the bottom of the Multiboot2 specification.
+#[repr(u32)]
+#[derive(Copy, Clone, Debug)]
+pub enum TagType {
+    End = 0,
+    Cmdline = 1,
+    BootLoaderName = 2,
+    Module = 3,
+    BasicMeminfo = 4,
+    Bootdev = 5,
+    Mmap = 6,
+    Vbe = 7,
+    Framebuffer = 8,
+    ElfSections = 9,
+    Apm = 10,
+    Efi32 = 11,
+    Efi64 = 12,
+    Smbios = 13,
+    /// Also called "AcpiOld" in other multiboot2 implementations.
+    AcpiV1 = 14,
+    /// Refers to version 2 and later of Acpi.
+    /// Also called "AcpiNew" in other multiboot2 implementations.
+    AcpiV2 = 15,
+    Network = 16,
+    EfiMmap = 17,
+    EfiBs = 18,
+    Efi32Ih = 19,
+    Efi64Ih = 20,
+    LoadBaseAddr = 21,
+}
+
+// each compare/equal direction must be implemented manually
+impl PartialEq<u32> for TagType {
+    fn eq(&self, other: &u32) -> bool {
+        *self as u32 == *other
+    }
+}
+
+// each compare/equal direction must be implemented manually
+impl PartialEq<TagType> for u32 {
+    fn eq(&self, other: &TagType) -> bool {
+        *self == *other as u32
+    }
+}
+
+impl PartialEq<TagType> for TagType {
+    fn eq(&self, other: &TagType) -> bool {
+        *self as u32 == *other as u32
+    }
+}
+
+impl PartialOrd<u32> for TagType {
+    fn partial_cmp(&self, other: &u32) -> Option<Ordering> {
+        let num = *self as u32;
+        Some(
+            if num < *other {
+                Ordering::Less
+            } else if num == *other {
+                Ordering::Equal
+            } else {
+                Ordering::Greater
+            }
+        )
+    }
+}
+
+/// All tags that could passed via the Multiboot2 information structure to a payload/program/kernel.
+/// Better not confuse this with the Multiboot2 header tags. They are something different.
+#[derive(Clone, Copy)]
 #[repr(C)]
 pub struct Tag {
-    pub typ: u32,
+    // u32 value
+    pub typ: TagType,
     pub size: u32,
     // tag specific fields
+}
+
+impl Debug for Tag {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Tag")
+            .field("typ", &self.typ)
+            .field("typ (numeric)", &(self.typ as u32))
+            .field("size", &(self.size))
+            .finish()
+    }
 }
 
 #[derive(Clone, Debug)]
@@ -28,7 +110,7 @@ impl<'a> Iterator for TagIter<'a> {
 
     fn next(&mut self) -> Option<&'a Tag> {
         match unsafe { &*self.current } {
-            &Tag { typ: 0, size: 8 } => None, // end tag
+            &Tag { typ: TagType::End, size: 8 } => None, // end tag
             tag => {
                 // go to next tag
                 let mut tag_addr = self.current as usize;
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -250,64 +250,39 @@ impl BootInformationInner {
...
-            for area in memory_map_tag.memory_areas() {
-                writeln!(
-                    f,
-                    "    S: {:#010X}, E: {:#010X}, L: {:#010X}",
-                    area.start_address(),
-                    area.end_address(),
-                    area.size()
-                )?;
-            }
+        for _t in  self.module_tags() {
```
IsaacWoods, line 256: `module_tags` returns an ~`impl Iterator` so this could just be `self.module_tags().count` I think

src/lib.rs
```diff
@@ -250,64 +250,39 @@ impl BootInformationInner {
...
-                writeln!(
-                    f,
-                    "    name: {:15}, S: {:#08X}, E: {:#08X}, L: {:#08X}, F: {:#04X}",
-                    s.name(),
-                    s.start_address(),
-                    s.start_address() + s.size(),
-                    s.size(),
-                    s.flags().bits()
-                )?;
+            for _s in elf_sections_tag.sections() {
```
IsaacWoods, line 261: Same use of `count` could be here I think

IsaacWoods (commented):
Just a few thoughts to start; I think this is probably an improvement. I'm away for the weekend but am happy to review after I get back.

src/lib.rs
```diff
@@ -250,64 +250,42 @@ impl BootInformationInner {
...
+                "command_line",
+                &self
+                    .command_line_tag()
+                    .map(|x| x.command_line())
+                    .unwrap_or(""),
+            )
+            .field("memory_areas", &self.memory_map_tag())
+            // so far, I didn't found a nice way to connect the iterator with ".field()" because
+            // the iterator isn't Debug
+            .field("module_tags (count)", &self.module_tags().count())
```
Caduser2020, line 274: I would prefer an implementation of Debug for the iterator if possible. In my experience, having module tags printed helps debugging. Also, you could try mapping over the iterator to Debug each element?

phip1611: I absolutely agree with you. The behaviour should be equal to `Vec::iter()` which also prints all elements. But with the current implementation of the ModuleIter I can't achieve this because of different compile time errors, like lifetime and ownership problems :/

phip1611:
Hi @IsaacWoods  @Caduser2020 
I tried to satisfy your review proposals. Unfortunately, I don't find a easy solution to print all elements when the iterator of for example ModuleTags is debug-formatted... there are problems with lifetimes and/or ownership all the time :/

I tried it to solve it like this:
```
impl <'a> Debug for ModuleIter<'a> {
    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
        let mut list = f.debug_list();
        self.iter.for_each(|ref e| {
            list.entry(e);
        });
        list.finish()
    }
}
``` 

but this results in compile errors... 
```
self.iter.for_each(|ref e| {
   |         ^^^^^^^^^ move occurs because `self.iter` has type `TagIter<'_>`, which does not implement the `Copy` trait
```

IsaacWoods:
Hmm, could you use `by_ref` to solve that? Like this (note, untested):
``` rust
impl <'a> Debug for ModuleIter<'a> {
    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
        let mut list = f.debug_list();
        self.iter.by_ref().for_each(|ref e| {
            list.entry(e);
        });
        list.finish()
    }
}
```

phip1611:
Nope, `by_ref` consumes `&mut self` which we don't have here.

IsaacWoods:
Might as well take the easy way out and `clone` it then, which [should work](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3924f76b347224a0b5d6f9c0b0e485b1) and is trivially cheap. Also note that by accessing `self.iter`, you're going to print all of the tags - `ModuleIter` implements `Iterator` itself and filters out the non-module tags. I've used the correct iterator in my Playground example.

phip1611:
Yes it worked with `.clone()` @IsaacWoods . In addition, I implemented an enum which helps us to get a nice name for each tag type. The enum is directly taken from the multiboot2 spec and I was really carefully with replacing existing numeric literals with enum variants. What do you think now?

Example output:
```
    Multiboot2 Boot Information {
        start_address: 0x000000000068e000,
        end_address: 0x00000000006f6960,
        total_size: 0x0000000000068960,
        boot_loader_name_tag: "GRUB 2.04-1ubuntu44.2",
        command_line: "",
        memory_areas: None,
        module_tags: [
            Tag {
                typ: 21,
                typ (name): MultibootTagTypeLoadBaseAddr,
                size: 12,
            },
            Tag {
                typ: 1,
                typ (name): MultibootTagTypeCmdline,
                size: 9,
            },
            Tag {
                typ: 2,
                typ (name): MultibootTagTypeBootLoaderName,
                size: 30,
            },
            Tag {
                typ: 9,
                typ (name): MultibootTagTypeElfSections,
                size: 428180,
            },
            Tag {
                typ: 12,
                typ (name): MultibootTagTypeEfi64,
                size: 16,
            },
            Tag {
                typ: 14,
                typ (name): MultibootTagTypeAcpiOld,
                size: 28,
            },
            Tag {
                typ: 15,
                typ (name): MultibootTagTypeAcpiNew,
                size: 44,
            },
            Tag {
                typ: 18,
                typ (name): MultibootTagTypeEfiBs,
                size: 8,
            },
            Tag {
                typ: 20,
                typ (name): MultibootTagTypeEfi64Ih,
                size: 16,
            },
        ],
        elf_sections_tags (count): 6689,
        efi_32_ih: None,
        efi_64_ih: Some(
            EFIImageHandle64 {
                typ: 20,
                size: 16,
                pointer: 114517528,
            },
        ),
        efi_sdt_32_tag: None,
        efi_sdt_64_tag: Some(
            EFISdt64 {
                typ: 12,
                size: 16,
                pointer: 127852568,
            },
        ),
        efi_memory_map_tag: None,
    }
```

src/header.rs
```diff
@@ -1,13 +1,90 @@
...
     pub typ: u32,
     pub size: u32,
     // tag specific fields
 }
 
+impl Debug for Tag {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.debug_struct("Tag")
+            .field("typ", &self.typ)
+            .field("typ (name)", unsafe { &(core::mem::transmute::<u32, TagType>(self.typ)) })
```
phip1611, line 82: any idea how this line can be written in a nicer way? I need to change the type from u32 to TagType

src/elf_sections.rs
```diff
@@ -96,6 +97,28 @@ impl Iterator for ElfSectionIter {
...
+            debug.entry(e);
+        });
+        debug.finish()
+    }
+}
+
+impl Default for ElfSectionIter {
+    fn default() -> Self {
+        Self {
+            current_section: 0 as *const u8,
```
Caduser2020, line 113: This and `string_section` below can be changed to use `core::ptr::null()`.

src/header.rs
```diff
@@ -1,13 +1,97 @@
 use core::marker::PhantomData;
+use core::fmt::{Debug, Formatter};
+use core::cmp::Ordering;
 
-#[derive(Clone, Copy, Debug)]
+/// Possible Types of a [`Tag`]. The names and values are taken from the example C code
+/// at the bottom of the Multiboot2 specification.
+#[repr(u32)]
+#[derive(Copy, Clone, Debug)]
+pub enum TagType {
```
IsaacWoods, line 9: Because we rely on the elements using the exact right indices, I think it would help to be more explicit here (e.g. `End = 0` for all of them). The comments are then not needed.

src/header.rs
```diff
@@ -1,13 +1,97 @@
...
+    Mmap,
+    Vbe,
+    Framebuffer,
+    ElfSections,
+    // 10
+    Apm,
+    Efi32,
+    Efi64,
+    Smbios,
+    AcpiOld,
```
IsaacWoods, line 27: As the maintainer of the `acpi` crates, I would probably prefer these to remain as `AcpiV1` and `AcpiV2` - I guess it can be argued that `V2` is confusing as it refers to later versions as well, but this would make it more consistent with the rest of the ecosystem.

src/lib.rs
```diff
@@ -250,64 +250,58 @@ impl BootInformationInner {
...
+            )
+            .field(
+                "command_line",
+                &self
+                    .command_line_tag()
+                    .map(|x| x.command_line())
+                    .unwrap_or(""),
+            )
+            .field("memory_areas", &self.memory_map_tag())
+            // so far, I didn't found a nice way to connect the iterator with ".field()" because
```
IsaacWoods, line 277: Outdated comment?

src/module.rs
```diff
@@ -51,7 +52,17 @@ impl<'a> Iterator for ModuleIter<'a> {
...
-            .find(|x| x.typ == 3)
+            .find(|x| x.typ == TagType::Module)
             .map(|tag| unsafe { &*(tag as *const Tag as *const ModuleTag) })
     }
 }
+
+impl <'a> Debug for ModuleIter<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        let mut list = f.debug_list();
+        self.clone().iter.for_each(|tag| {
```
IsaacWoods, line 63:
```suggestion
        self.clone().for_each(|tag| {
```
You shouldn't call through to the inner `iter` here - this will print all of the tags, as shown in your example output, instead of just the module ones. `self` also implements `Iterator`, and correctly filters out just the module tags.

IsaacWoods (changes requested):
Thanks for working on this - great changes overall I think, and I like the explicit enum for tag types. Just a few straggling changes left.

As one last suggestion, before we merge, it might be nice to squash into fewer commits that collect changes to specific areas together. Are you comfortable performing an interactive rebase or would you prefer me to do this?

phip1611:
> As one last suggestion, before we merge, it might be nice to squash into fewer commits that collect changes to specific areas together. Are you comfortable performing an interactive rebase or would you prefer me to do this?

No problem, I'm well familiar with this.

I adjusted everything. Hope I didn't miss anything.

IsaacWoods (approved):

IsaacWoods:
Great - thanks @phip1611 for this work! I'll merge this and get it published when I get the chance.

PR #77, opened 2021-07-15
add multiboot2 magic number

phip1611:
To my surprise the multiboot2 magic number is not yet available inside the crate. In my Rust kernel I currently have a hard-coded constant. I think it would be beneficial, if we upstream this into the crate.


IsaacWoods:
Thanks!

PR #79, opened 2021-07-17
Restructure library into "mbi" and "header"-part + add header parsing

phip1611:
This is a huge PR, that adds lots of functionality but doesn't change existing code except some restructuring. Multiboot2 consists of two parts. I refer to them as "mbi" and "header". The mbi-part is everything that was covered so far: parsing the multiboot2 information structure (mbi). **I added (~1000 lines) to also parse and construct multiboot2 headers with this PR.** It includes two examples.

- [x] make a clear difference between `MbiTags` and `HeaderTags`
- [x] add all necessary code  to parse multiboot2 headers
- [x] add examples to create and test parsing of multiboot2 headers 
- [x] split library into "mbi" and "header"-part
- [ ] refactor new code so that it returns more results and doesn't panic if something goes wrong (+ error type)
- [ ] more comments + module documentation
- [ ] discuss breaking changes: with smart imports we could do everything without  any breaking change (but I'd recommend  and release `v1.0.0` soon, which could contain breaking changes (we will have them anyway, see other PRs))

(The following points are created from suggestions made in the review. Because the PR is so big, I think it is useful to capture them here)
- [ ] remove alloc crate again (which was a mistake) + rethink builder-feature
- [ ] implement more convenient getters on `Multiboot2Header` (for each tag type)
- [ ] move header-load-function to `Multiboot2Header::`
- [ ] README should reflect the new changes
- [ ] CHANGELOG!
- [ ] Cargo.toml description (+ keywords, +category? new issue?)


phip1611:
@IsaacWoods  seems I still need reviews, even tho I'm part of the organisation now. Please look first into the edited PR description.. and sorry for the BFR of code to review (+2500, −1,391) (**The PR is almost only additions. The "minus" lines are just a moving of the existing files into a sub directory.**)

In short all existing code is now in `mbi` module and I added the `headers` module.

I was really careful while implementing the header-parsing and controlled everything against the spec multiple times. Any opinion so far? I think this will make the crate ready for a `v1.0.0` release soon.

README.md
```diff
@@ -1,11 +1,20 @@
...
+the tags from the "multiboot2 information structure" (also called *mbi*). This is the data structure passed 
+via `ebx` register (on `x86`) when the bootloader gives control to the payload (for example your kernel).
+Furthermore, it contains type definitions for parsing a "multiboot2 header" including the "multiboot2 header tags".
 
-It uses the Multiboot 2.0 specification at https://www.gnu.org/software/grub/manual/multiboot2/multiboot.html and the ELF 64 specification at http://www.uclibc.org/docs/elf-64-gen.pdf.
+It's still incomplete, so please open an issue if you're missing some functionality. Contributions welcome!
+It uses the Multiboot 2.0 specification at https://www.gnu.org/software/grub/manual/multiboot2/multiboot.html and the 
+ELF 64 specification at http://www.uclibc.org/docs/elf-64-gen.pdf.
+
+⚠ The `elf64` in the github repository url is a bit misleading. It means the crate can parse the `elf`-sections
```
IsaacWoods, line 15: I probably agree with this, but rather than add this comment, I think it would be better to rename the repository. The crate is already just called `multiboot2`

cc @rust-osdev/multiboot2 does anyone feel that we shouldn't do that?

examples/parse_mb2_header_runtime.rs
```diff
@@ -0,0 +1,76 @@
+extern crate multiboot2;
```
IsaacWoods, line 1: I'm not sure this is a great example of what users should be doing, and doesn't really showcase this crate. I think we should just stick with the other example.

examples/parse_mb2_header_static.rs
```diff
@@ -0,0 +1,30 @@
+use std::fs::File;
+use std::io::Read;
+
+/// Reads a multiboot2 header that is stored in a as binary in a file.
+fn main() {
+    let mut bin = File::open("examples/mb2_header.bin").unwrap();
+    let mut data = Vec::new();
+    bin.read_to_end(&mut data).unwrap();
+    let data_ptr = data.as_ptr();
```
IsaacWoods, line 12: Maybe the `include_bytes!` macro would be better for an example?

src/lib.rs
```diff
@@ -30,19 +28,21 @@ pub use vbe_info::{
     VBECapabilities, VBEControlInfo, VBEDirectColorAttributes, VBEField, VBEInfoTag,
     VBEMemoryModel, VBEModeAttributes, VBEModeInfo, VBEWindowAttributes,
 };
+use tags::mbi::{TagType, Tag, TagIter};
 
 #[macro_use]
 extern crate bitflags;
+extern crate alloc;
```
IsaacWoods, line 13: I don't think we can do this - people need to parse the multiboot structures before they have a heap set up, generally. This actually caused me to check and we actually need to move this crate over to the 2018 edition... oops.

src/tags/header.rs
```diff
@@ -0,0 +1,970 @@
...
+use core::fmt::{Debug, Formatter};
+use core::mem::size_of;
+use tags::mbi::TagType;
+use core::marker::PhantomData;
+
+
+/// Value must be present in multiboot2 header.
+pub const MULTIBOOT2_HEADER_MAGIC: u32 = 0xe85250d6;
+
+pub unsafe fn load_mb2_header<'a>(addr: usize) -> Multiboot2Header<'a> {
```
IsaacWoods: Why not as a method on `Multiboot2Header`?

IsaacWoods (commented):
Blimey - this is a huge PR with a large amount of new functionality we hadn't necessarily planned to provide. This is probably more attention that this library has gotten in quite a while, and unfortunately it does show the bitrot a little. For a start, I hadn't realised that the library was still on the `2015` edition - we should definitely migrate it over at some point.

I've added a few comments for now, but it's pretty difficult to review this PR with the multiple changes to structure on top of all of the new code. I'll need to come back to this with a brain a little less foggy. It might be easier to split out some of the smaller changes for now, and then discuss the organisation of the header structures a bit more, so it can be done in one history "jump". What do you think?

To start, what is the motivation for providing the header parsing stuff? I think the only projects interested in doing that would be bootloaders - is that the idea, or is this just for completeness?

---

>  seems I still need reviews, even tho I'm part of the organisation now.

Yes, that is the arrangement we have on this repo. Being part of the organisation allows you to push to this repo directly, but that should not be used for pushing features, and all PRs go through code review. FWIW, it would not have been appropriate to merge this without review.

examples/parse_mb2_header_runtime.rs
phip1611: First: The `extern crate multiboot2` was accidentally added by my IDE. Second: I think it would be better to move it to `tests` directory?

src/lib.rs
phip1611: Ah yeah, this was my IDE during development. It is not necessary anymore and can be removed.

src/tags/header.rs
phip1611: I wanted to structure the library similar to the existing code. But yeah, good idea.

phip1611:
Thanks for reviewing this! I know the PR is huge but I also didn't know if it would be easy to structure it into multiple PRs in a sensible way.. because the changes belong together.

> To start, what is the motivation for providing the header parsing stuff? I think the only projects interested in doing that would be bootloaders - is that the idea, or is this just for completeness?

Both and additionally, I'd love if this crate becomes a good documentational source about how `multiboot2` works. The original spec isn't that good. But mainly for feature completeness, so that eventually someone can write GRUB or so in Rust and parse multiboot2 headers.

>  For a start, I hadn't realised that the library was still on the 2015 edition - we should definitely migrate it over at some point.

Yikes, that's true. I created #83  to keep track of this.

Maybe we add a milestone `v1.0.0` and add the relevant issues/PRs to it?

robert-w-gries:
@IsaacWoods I can help review if you want, but I need time to go through the library and refresh my memory on how the internals work.

@phip1611 Thanks for your interest in the project!

I skimmed through the PR and noticed breaking changes to the API. I'm generally okay with adding new (safe) public items, but I tend to be more careful when it comes to breaking changes for users. I think it's important to distinguish between the `HeaderTag`s and `MbiTag`s, but if there's no way to avoid breaking changes then I think we'll need to strategize on how to add these changes without disrupting users.

I'd like to consider the following before merging in a feature as large as this:

1. Providing a release with clean up and fixes (if any) so users have a good library version to use if they don't want to migrate to `v1.0` or whatever version this would be
2. Moving the `examples` to `tests` and adding back the tests that were removed from `lib.rs`
3. Agreeing on a clear direction for the library and notifying users so they can decide whether the new version is for them

I'm open to discussion for the above points, and I think @IsaacWoods is more qualified than me regarding the direction of the library.

Finally, I want to add that it's great to see the library being driven forward. I love the Rust OSDev ecosystem, and it's good to see people still hacking on it and improving it. It makes me want to start working on my kernel again 😄

IsaacWoods:
Good to hear from you @robert-w-gries! A second pair of eyes would be great if you have the time.

> Finally, I want to add that it's great to see the library being driven forward. I love the Rust OSDev ecosystem, and it's good to see people still hacking on it and improving it. It makes me want to start working on my kernel again

Seconded! I don't personally use a multiboot2 loader for my OS anymore, but I think it's a fairly fundamental part of the OSdev ecosystem, especially for newer people, so it's great to see some interest being taken in it.

---

+1 to Rob's plan - I think filtering improvements down (e.g. moving to the new edition, #80 which I think is fundamentally a good idea, etc.) before doing the header stuff is the way to go.

Organisationally wise, I wonder if a separate crate for the header stuff (e.g. `multiboot2-header`) would be better. This would nicely separate the header and boot info structures, and allow users implementing a kernel to just import the stuff they wanted. Bootloaders could then just use both crates if needed. This is an approach we've used for [the ACPI crates](https://github.com/rust-osdev/acpi) with (I think) good results. It would also allow for the header stuff to be published and versioned separately from the existing crate (I think adding all that new code would set back stabilisation until it had time to bake, fwiw).

> Both and additionally, I'd love if this crate becomes a good resource about how multiboot2 works. The original spec isn't that good or specific enough at a first glance. But mainly for feature completeness, so that eventually someone can write a bootloader or so in Rust and parse multiboot2 headers.

Ah fair enough, I agree the multiboot spec could be a lot better (and have thought about writing writing out a nicer version for this crate, but have never gotten around to it [feel free to take the torch on that one ;) ]). I think providing types for the headers is a good idea and would make writing a multiboot parser that much easier, but am conscious of adding many types and confusing notions of multiple sets of tags to this crate. What do you folks think about the multiple crates idea?

phip1611:
To speed things up I'd like to come to the most important point first.

> What do you folks think about the multiple crates idea?

I'm in, good idea @IsaacWoods @robert-w-gries Let's do it.
I'm not exactly sure how to create a new `multiboot2-header` repo in the `rust-osdev` space. Can somebody of you do this?

Then I'd move my code to the repository and we shut down this PR without changes to not overload it anymore. Does anyone aggree?

The new `multiboot2-header`-crate would need to depend on the `multiboot2`-crate, because I need `enum [Mbi]TagType` in `multiboot2-header`. For this, I'd make changes in this crate (in new, fresh PRs), that

- bump this to the rust edition 2018
- make `load` return Result
- publicly export `enum TagType`

I'd like to release this as `multiboot2 @ v1.0.0` because this crate is no longer "experimental" and production ready.

Do we have a clear plan now?

IsaacWoods:
> I'm not exactly sure how to create a new multiboot2-header repo in the rust-osdev space. Can somebody of you do this?

We don't need a separate repository. We would turn this repository into a Cargo workspace (this would require us to move the current crate into a sub-directory, `multiboot2`), and then add another crate alongside it. Look at the [acpi](https://github.com/rust-osdev/acpi) for an example. Creating the workspace will create a bit of renaming churn, so probably is best to do as a PR of its own, before adding the header crate.

> I'd like to release this as multiboot2 @ v1.0.0 because this crate is no longer "experimental" and production ready

I'm not sure I agree with this - releasing a v1 of a crate is not a magic mark that it is not "experimental" (I agree with removing that labeling though), but does reduce our ability to create breakage. For example, I think I released a stable version of `acpi` too soon, and it's made changing stuff that much more painful. In due course a stable version can be released, but I don't think it's a good idea to do so while you're planning such large changes.

phip1611:
Yeah sounds reasonible. So if #84, #80 and #75 are resolved/merged, I'd create a PR for a cargo workspace.

I worked with workspaces before, no problem. Good idea!

IsaacWoods:
> Does this sound like a good plan?

Good plan - I've merged or left comments on those three PRs. If the original author of #75 doesn't come back to us, we can just rewrite the commit with the new method names and add them as a co-author.

> Hm but we'll have a breaking change anyway

Exactly :) while we have low-hanging fruit like that that requires breakage, imo it's better to keep the crate unstable. We may well need additional changes in the near future - you can quite quickly end up burning through major versions for small breaking changes, which imo is against the spirit of stable major versions, and can create an impression of poor maintainership. I think it'll be better to publish these changes as `v0.12.0` for now - it's still fine to plan towards a stable release (in e.g. a few months) if you'd like.

phip1611:
Closed in favor of #95

PR #80, opened 2021-07-17
load returns result (no more assertions that could panic)

phip1611:
(Before merging I'd recommend to wait for https://github.com/rust-osdev/multiboot2-elf64/pull/79)

Breaking Change: `load()` returns now a Result. Additionally, a new `MbiLoadError`-enum was introduced.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,7 +46,7 @@ mod vbe_info;
 
 /// Load the multiboot boot information struct from an address.
 ///
-/// This is the same as `load_with_offset` but the offset is ommitted and set
+/// This is the same as `load_with_offset` but the offset is omitted and set
 /// to zero.
 ///
 /// Examples
@@ -55,19 +55,12 @@ mod vbe_info;
 /// use multiboot::load;
 ///
 /// fn kmain(multiboot_info_ptr: u32) {
-///     let boot_info = load(ptr as usize);
+///     let boot_info = load(ptr as usize).unwrap();
 ///     println!("{:?}", boot_info);
 /// }
 /// ```
-pub unsafe fn load(address: usize) -> BootInformation {
-    assert_eq!(0, address & 0b111);
-    let multiboot = &*(address as *const BootInformationInner);
-    assert_eq!(0, multiboot.total_size & 0b111);
-    assert!(multiboot.has_valid_end_tag());
-    BootInformation {
-        inner: multiboot,
-        offset: 0,
-    }
+pub unsafe fn load(address: usize) -> Result<BootInformation, MbiLoadError> {
+    load_with_offset(address, 0)
 }
 
 /// Load the multiboot boot information struct from an address at an offset.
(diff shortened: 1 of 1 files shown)


phip1611:
Ready for review @IsaacWoods  :)

src/error.rs
```diff
@@ -0,0 +1,33 @@
...
+    IllegalAddress,
+    /// The total size of the multiboot2 information structure must be a multiple of 8.
+    /// (Not in spec, but it is implicitly the case, because the begin of MBI
+    /// and all tags are 8-byte aligned and the end tag is exactly 8 byte long).
+    IllegalTotalSize(u32),
+    /// End tag missing. Each multiboot2 header requires to have an end tag.
+    NoEndTag,
+}
+
+impl Display for MbiLoadError {
```
IsaacWoods, line 19: Hmm, I'm not sure implementing `Display` for errors like this adds much over the `Debug` implementation

src/lib.rs
```diff
@@ -37,6 +38,7 @@ extern crate bitflags;
 mod boot_loader_name;
 mod command_line;
 mod elf_sections;
+mod error;
```
IsaacWoods, line 41: This doesn't really need to be in a whole module, and leaving it closer to the function it's produced by helps readability imo

src/lib.rs
```diff
@@ -59,14 +61,28 @@ mod vbe_info;
...
     let multiboot = &*(address as *const BootInformationInner);
-    assert_eq!(0, multiboot.total_size & 0b111);
-    assert!(multiboot.has_valid_end_tag());
-    BootInformation {
-        inner: multiboot,
-        offset: 0,
+
+    // Check if total size is a multiple of 8.
+    // See MbiLoadError::IllegalTotalSize for comments
+    if 0 != multiboot.total_size & 0b111 {
```
IsaacWoods, line 74: Are these... Yoda conditions? These are not needed in Rust as it has sensible type checking on `if`, and obfuscate the code.

src/lib.rs
```diff
@@ -59,14 +61,28 @@ mod vbe_info;
...
+        return Err(MbiLoadError::IllegalTotalSize(multiboot.total_size));
+    }
+    if multiboot.has_valid_end_tag() {
```
IsaacWoods, line 77: Might be cleaner to handle all of the invalid conditions with separate `if`s, and then leave the successful outcome as the default (i.e handle `!multiboot.has_valid_end_tag`, and then just return the `Ok` if you haven't returned yet).

src/error.rs
phip1611: I don't mind if we don't have it. BTW: My general rule is that if in doubt to implement a little more than a little to less. There is no disadvantage in having it. @IsaacWoods

src/lib.rs
phip1611: I reused code of existing conditions. But I agree that they aren't pretty and can be optimized.

src/lib.rs
```diff
@@ -55,19 +55,12 @@ mod vbe_info;
...
-    assert_eq!(0, address & 0b111);
-    let multiboot = &*(address as *const BootInformationInner);
-    assert_eq!(0, multiboot.total_size & 0b111);
-    assert!(multiboot.has_valid_end_tag());
-    BootInformation {
-        inner: multiboot,
-        offset: 0,
-    }
+pub unsafe fn load(address: usize) -> Result<BootInformation, MbiLoadError> {
+    load_with_offset(address, 0)
```
phip1611, line 63: much cleaner now. Previously, both functions used the same code twice to check everything -> removed code duplication.

IsaacWoods:
Cool, I think this is definitely an improvement on the situation before. Thanks!

Looks like this needs a rebase to resolve merge conflicts, and then I can merge!

phip1611:
@IsaacWoods ready to merge

IsaacWoods (approved):

PR #82, opened 2021-07-17
Fixing future compiler error "unaligned_references" (82523)

phip1611:
Fixing a future compile error, that is currently a warning. More info: https://github.com/rust-lang/rust/issues/82523

The  fix with curly-syntax is quite simple. It creates a new copy of the values, iff the are `copy`. Easy solution.


src/vbe_info.rs
```diff
@@ -86,20 +86,18 @@ pub struct VBEControlInfo {
...
-                .field("total_memory", &self.total_memory)
-                .field("oem_software_revision", &self.oem_software_revision)
-                .field("oem_vendor_name_ptr", &self.oem_vendor_name_ptr)
-                .field("oem_product_name_ptr", &self.oem_product_name_ptr)
-                .field("oem_product_revision_ptr", &self.oem_product_revision_ptr)
-                .finish()
-        }
+        f.debug_struct("VBEControlInfo")
+            .field("signature", &self.signature)
+            .field("version", &{self.version})
```
phip1611, line 91: this curly syntax creates a new copy, because the value itself is `copy`

IsaacWoods (approved):

IsaacWoods:
Thanks!

PR #86, opened 2021-07-28
prepared cargo workspace, as discussed in PR #79

phip1611:
As discussed in #79 , this PR prepares the cargo workspace.

*Postpone until #90  and #89  are merged.*


multiboot2-mbi/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
phip1611, line 1: Hint: Same content as the original file. No changes here.

IsaacWoods:
Great, these small steps are a lot easier to review. Thanks for your patience!

I do wonder if renaming the current crate to `multiboot2-mbi` is confusing, especially as the actual package remains `multiboot2`. What do you think of keeping the current crate as `multiboot2` and then adding the new `multiboot2-header` alongside it?

Is there anything left you'd like to do for `v0.12.0`, or should we publish that so we can get this merged?

phip1611:
> I do wonder if renaming the current crate to multiboot2-mbi is confusing, especially as the actual package remains multiboot2. What do you think of keeping the current crate as multiboot2 and then adding the new multiboot2-header alongside it?

We can leave the name as it.

> Is there anything left you'd like to do for v0.12.0, or should we publish that so we can get this merged?

I think only #87 is missing.

multiboot2-header/src/lib.rs
```diff
@@ -0,0 +1 @@
+//! TODO
```
phip1611, line 1: Part of another PR

multiboot2/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
```
phip1611, line 1: Nothing changed, just moved the file

phip1611:
Are we ready to go, @IsaacWoods ?

multiboot2-header/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "multiboot2-header"
+description = """
+Library with type definitions and parsing functions for Multiboot2 headers.
+This library is `no_std` and can be used in a bootloaders.
+"""
+# TODO keep in Sync with "multiboot2"
+version = "0.12.0"
```
IsaacWoods, line 8: In my experience with the ACPI crates, this is not the best idea, as it confuses people into thinking the crate has changed when it has not, and doesn't really have any advantages imo. I think we should let them float at separate versions. This would also allow one to be stabilized separately from the other.

phip1611: Thanks, I wasn't sure about that! :)

IsaacWoods (approved):

IsaacWoods:
Great, I think this is good to go @phip1611? I'll leave it to you to merge as I guess you'll need to publish the new crate. It would probably be good to add the `rust-osdev/multiboot2` team as an owner so we can publish fixes and whatnot if needed.

PR #87, opened 2021-08-05
Cargo toml prepare relase v0.12 + changelog

phip1611:
Hey @IsaacWoods 
what do you think? If we are ready to go, we can merge this and afterwards focus on #86 

- adjusted Cargo.toml
- adjusted lib.rs documentation
- adjusted README
- adjusted Changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,34 @@
 [package]
 name = "multiboot2"
-version = "0.11.0"
-authors = ["Philipp Oppermann <dev@phil-opp.com>", "Calvin Lee <cyrus296@gmail.com>", "Isaac Woods"]
+description = """
+Library that helps you to parse the multiboot information structure (mbi) from
+Multiboot2-compliant bootloaders, like GRUB. It supports all tags from the specification
+including full support for the sections of ELF-64. This library is `no_std` and can be
+used in a Multiboot2-kernel.
+"""
+version = "0.12.0"
+authors = [
+    "Philipp Oppermann <dev@phil-opp.com>",
+    "Calvin Lee <cyrus296@gmail.com>",
+    "Isaac Woods",
+    "Philipp Schuster <phip1611@gmail.com>"
+]
 license = "MIT/Apache-2.0"
-description = "An experimental Multiboot 2 crate for ELF-64/32 kernels."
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -1,11 +1,34 @@
...
+]
+keywords = [
+    "Multiboot2",
+    "kernel",
+    "boot",
+]
+# without this, sometimes crates.io doesn't show the preview of the README
+# I expeciended this multiple times in the past
+readme = "README.md"
+homepage = "https://github.com/rust-osdev/multiboot2-elf64"
```
IsaacWoods, line 29: Did we decide to rename the repo or not in the end? If we do (I'm probably in favour of going ahead with renaming to just `multiboot2`), we want to leave these as is.

phip1611: We can! Because this will cause in broken git urls anyway, do we want to use this opportunity to rename `master` branch to `main` too? 

Can I do this with my rights? Not sure, I'm on phone right now. 🤔 

I'd still keep the (updated) URLs for github repository, homepage and repository, just for completeness (I always do this, I'm kind of a perfectionist). There is no disadvantage, right?

IsaacWoods: Team members have the `write` role, which I don't think is enough to manage repository settings - as this won't be a regular thing, I'll just make the changes as maintainer for simplicity.

Apparently Github will automatically redirect URLs referring to the old repo and branch name, which is nice and won't break anything for now. Contributors should probably update their remotes anyway (cc @rust-osdev/multiboot2).

> I'd still keep the (updated) URLs for github repository, homepage and repository in Cargo.toml, just for completeness (I always do this, I'm kind of a perfectionist). There is no disadvantage, right? But I do not insist on that..

Ah no that's fine - I just meant we should update these to the updated URLs if we do the rename, not that we shouldn't add these.

phip1611:
URLs fixed, great! :) Any more comments on this PR?

IsaacWoods (approved):

IsaacWoods:
Looks good to me - thanks for working on this :)

I've approved the PR. You should have permissions to do a `cargo publish` - if that goes through successfully, feel free to merge this.

phip1611:
Wonderful, worked. `v0.12` is released :)

PR #88, opened 2021-08-06
renamed old Github urls in README

phip1611:
Seems like we missed to change a few URLs. Thanks to URL redirecting from Github it's not critical.

PS: I thought you said I'd be able to push small things directly to `main` without a review. This is not possible with my current rights. I'm fine with PRs, but wanted you to know this.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,9 +18,9 @@
 
 # 0.9.0
 
-- Add a `checksum_is_valid` method to the RSDP tags ([#64](https://github.com/rust-osdev/multiboot2-elf64/pull/64))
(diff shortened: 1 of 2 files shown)


IsaacWoods:
Ah thanks, I missed these.

> PS: I thought you said I'd be able to push small things directly to main without a review. This is not possible with my current rights. I'm fine with PRs, but wanted you to know this.

Hm, sorry about that - you should be able to. I think this is due to a branch protection rule which didn't affect me as `admin` but does as the team is `write`. I've moved the team to the `maintain` role - I think that should fix this issue.

IsaacWoods (approved):

PR #89, opened 2021-08-11
v0.12.1 Unpublic tags bugfix

phip1611:
Already published on crates.io because it is a minor adjustment. I still am not able to push small things to `main`, therefore I created the PR after the release on crates.io.

(as we discussed here: https://github.com/rust-osdev/multiboot2/pull/88)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ Multiboot2-compliant bootloaders, like GRUB. It supports all tags from the speci
 including full support for the sections of ELF-64. This library is `no_std` and can be
 used in a Multiboot2-kernel.
 """
-version = "0.12.0"
+version = "0.12.1"
 authors = [
     "Philipp Oppermann <dev@phil-opp.com>",
     "Calvin Lee <cyrus296@gmail.com>",
@@ -16,6 +16,7 @@ authors = [
 license = "MIT/Apache-2.0"
 edition = "2018"
 categories = [
+    "no-std",
     "parsing",
 ]
 keywords = [
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,7 @@
+# 0.12.1
+- `TagType`-enum introduced in `v0.11` is now actually public
+- internal code improvements
+
 # 0.12.0
 
 - **breaking:** `load()` and `load_with_offset` now returns a result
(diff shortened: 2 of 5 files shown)


IsaacWoods:
Looks good to me, thanks!

It's really weird that you still can't push - I've added the multiboot2 team explicitly to the branch protection rule, but since you have the Maintain role, that should already work. If this doesn't fix it, I guess I'll have to remove the rule, but I don't want to do that unless necessary.

(cc @rust-osdev/multiboot2, if anyone has time, could someone else try pushing to `main` [feel free to then just delete the commit and force-push if its done within a few minutes]).

This is what the rule currently looks like - maybe I'm misunderstanding:
image

phip1611:
> I guess I'll have to remove the rule, but I don't want to do that unless necessary.

We should keep the rule. But yeah, that's strange

phil-opp:
I think the relevant part of the rule description is _"Required status checks will still prevent these people, teams and apps from merging if the checks fail."_ As I understand it, this rule allows to further restrict who can push to the branch instead of allowing pushes from all people that have write access.

But: You can still only push commits that meet all the other requirements too. This means that the commit that you want to push must have been approved and the CI checks must pass. For example, you could create a PR, wait for CI pass and approval, and then close the PR and push the approved commit directly. This doesn't make much sense of course :D.

The only people that can ignore the rules are those with the admin role. There is a separate switch to disable this if desired, but it is not possible relax it further for other roles:
image

phip1611:
> This doesn't make much sense of course :D.
If it works.. :D I would only do this for minor stuff anyway

Thanks for the hint, Philipp!

phil-opp:
The problem is that you still would need to create a PR and wait for a review. The only difference is that instead of clicking the merge button you push directly.

phip1611:
Ah okay, no problem. Then I stick with PRs.. As long as @IsaacWoods  has enough time for the reviews, I'm fine  🤣

IsaacWoods:
> I think the relevant part of the rule description is "Required status checks will still prevent these people, teams and apps from merging if the checks fail." As I understand it, this rule allows to further restrict who can push to the branch instead of allowing pushes from all people that have write access.

Huh, okay, thanks for clarifying that Phil. I read it as meaning that the named people or teams could push the same way as administrators, but that PRs would still need to pass status checks. Looks like they've changed this recently (users with the Maintain role used to be able to push, but now the text has changed):
image
image

I'm fine reviewing changes, but I don't want to seem like the sole gatekeeper of the `main` branch; I trust the team and y'all should be able to push :P

For now, I think I will remove the branch protection rule, but I would ask us to maintain explicit code review + approval and a CI pass before merging as a social expectation of sorts. We're a small enough team of maintainers that I think this is feasible, but I do wish I could keep the formal rule but exempt people of the Maintain role.

PR #90, opened 2021-08-11
Stricter typing + bugfix 

phip1611:
*[NEEDS TO BE REBASED, AFTER #89 GETS MERGED]*

This PR introduces stricter typing and a bugfix (wrong enum variant values).


src/memory_map.rs
```diff
@@ -67,33 +68,32 @@ impl MemoryArea {
...
+    /// A reserved area that must not be used.
+    Reserved = 2,
 
     /// Usable memory holding ACPI information.
-    AcpiAvailable,
+    AcpiAvailable = 3,
 
     /// Reserved memory which needs to be preserved on hibernation.
-    ReservedHibernate,
+    /// Also called NVS in spec, which stands for "Non-Volatile Sleep/Storage",
```
IsaacWoods, line 91: Good to mention it's called `NVS` in the C code, but I think we should preserve the bit about this memory needing to be preserved when hibernating.

IsaacWoods:
Good catch on the enum variant. One nit about a comment but then feel free to rebase and merge.

phip1611:
I rebased everything @IsaacWoods 

> but I think we should preserve the bit about this memory needing to be preserved when hibernating.

I'm sorry, I don't know what you mean. I didn't removed anything, only made an addition?

IsaacWoods (approved):
> I'm sorry, I don't know what you mean. I didn't removed anything, only made an addition?

Sorry, brainfart on my part. I skimmed over it and saw the original comment marked for removal I guess.

PR #92, opened 2021-08-12
Code style improvements + optional CI job (clippy, rustfmt, rustdoc)

phip1611:
- add clippy check as required
- add rustfmt check as required
- add rustdoc check as required (`#![deny(rustdoc::all)]` works since stable 1.52.0)

To improve and enforce code quality.

I'm not sure if we still use Travis CI. 

What's your opinion on this @Caduser2020  @IsaacWoods ?

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -11,11 +11,10 @@ use core::marker::PhantomData;
 pub const MULTIBOOT2_BOOTLOADER_MAGIC: u32 = 0x36d76289;
 
 /// Possible types of a Tag in the Multiboot2 Information Structure (MBI), therefore the value
-/// of the the `typ` field in [`Tag`]. The names and values are taken from the example C code
+/// of the the `typ` property. The names and values are taken from the example C code
 /// at the bottom of the Multiboot2 specification.
 #[repr(u32)]
 #[derive(Copy, Clone, Debug)]
-#[allow(missing_docs)]
 pub enum TagType {
     /// Marks the end of the tags.
     End = 0,
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,3 +20,18 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+
+
+  # As discussed, these tasks are optional for PRs.
+  style_checks:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Rustfmt
+      run: cargo fmt -- --check
+    - name: Clippy
+      run: cargo clippy
+    - name: Rustdoc
+      run: cargo doc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,0 @@
-language: rust
-rust:
-  - stable
-  - beta
-  - nightly
-matrix:
-  allow_failures:
-    - rust: stable
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,14 @@
 #![no_std]
 #![deny(missing_debug_implementations)]
-#![deny(missing_docs)]
+// --- BEGIN STYLE CHECKS ---
+// These checks are optional in CI for PRs, as discussed in
+// https://github.com/rust-osdev/multiboot2/pull/92
+#![deny(clippy::all)]
+#![deny(rustdoc::all)]
+// Forcing this would be a little bit ridiculous, because it would require code examples for
+// each getter and each trivial trait implementation (like Debug).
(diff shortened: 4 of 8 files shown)


src/header.rs
```diff
@@ -11,11 +11,10 @@ use core::marker::PhantomData;
 pub const MULTIBOOT2_BOOTLOADER_MAGIC: u32 = 0x36d76289;
 
 /// Possible types of a Tag in the Multiboot2 Information Structure (MBI), therefore the value
-/// of the the `typ` field in [`Tag`]. The names and values are taken from the example C code
```
phip1611, line 14: `Tag` is private, therefore rustdoc failure

IsaacWoods:
Hmm, I've tried having these sorts of checks in CI with communal projects before, and I think I'm generally against it. It makes merging every PR, especially from new contributors, a bit of an uphill battle that doesn't leave a good impression for potential contributors. I think with a relatively small project, doing a clippy PR every so often, as well as manually asking for really bad formatting to be run through rustfmt before you merge is good enough. Open to alternative views though.

> I'm not sure if we still use Travis CI.

Ah no, my mistake. Afaik, Travis CI is basically dead (it stopped reporting status updates for this repo ages ago). I fixed that by moving us oven to Actions, but must've forgotten to delete the Travis files. Feel free to remove them :)

phip1611:
@IsaacWoods 

>  It makes merging every PR, especially from new contributors, a bit of an uphill battle that doesn't leave a good impression for potential contributors

Hm. On the one hand I can understand this. On the otherhand, I'm certain that contributors of Multboot2, which is not a super basic domain but deep into OS development, will also not be total Rust newbees. Furthermore, it's a good opportunity for them to learn to use `rust fmt` and `rust clippy`. 

Any other opinion? @Caduser2020  @phil-opp

phip1611:
Any update here @Caduser2020  @phil-opp  @IsaacWoods ?

IsaacWoods:
I stand by my view that this has more disadvantages than advantages, and that CI failures for slight differences in code style are generally unhelpful (this is especially bad in OSdev when lots of people are a few nightlies behind, and so `rustfmt` complains not because you didn't format, but you formatted with the wrong version of it). Lots of people's editors autoformat, and I'd rather ask for an individual PR to be reformatted if I felt that the style was confusing or egregious. I also think that it generally annoys people, newbees and otherwise, and most people become exposed to those tools anyway.

So my vote is to close, but I'm also not going to make this a hill to die on, so if you feel particularly strongly, you can merge.

phip1611:
> I stand by my view that this has more disadvantages than advantages, and that CI failures for slight differences in code style are generally unhelpful (this is especially bad in OSdev when lots of people are a few nightlies behind, and so rustfmt complains not because you didn't format, but you formatted with the wrong version of it

Theoretically, we could use a "rust-toolchain.toml" for that :)

But I do not insist on the CI requirement. I like it for private projects but will remove it here. Will keep the formatting and style improvements, tho.

phil-opp:
> Any other opinion? @phil-opp

First of all, I'm not maintaining the `multiboot2` crate anymore, so my opinion on this doesn't really matter :D.

I don't think that it's a good idea to do add `#![deny(clippy::all)]` or `#![deny(rustdoc::all)]` to the code. The problem is that new clippy/rustdoc versions sometimes introduce new warnings, which would then break everyone who depends on this library. It's better to pass the deny argument on the command line in the CI script.

My personal preference is to keep warnings as warnings and use e.g. https://github.com/actions-rs/cargo to show the warnings in the "Files Changed" tab on GitHub. This way, the maintainers can decide whether the warning should be fixed or not, and it doesn't affect the mergeability of newcomer PRs.

Enforcing rustfmt is ok in my opinion, but I agree with @IsaacWoods that doing it on a crate that targets both stable and nightly can lead to problems. If you decide to add such a check, I would again recommend to create a separate CI job for it, which is not required to pass. This way, formatting problems don't block merging of newcomer PRs (they should still be fixed in a follow-up PR though).

phip1611:
>  I would again recommend to create a separate CI job for it, which is not required to pass. This way, formatting problems don't block merging of newcomer PRs (they should still be fixed in a follow-up PR though).

I agree, let's do it that way :)

Caduser2020:
I agree with @IsaacWoods for this. My build of rust is often a few days behind and rustfmt would become out of sync. I am okay with fixes for these warnings, but breaking CI is not worth it. Vote to close.

phip1611:
After all, this commit/PR 
- introduces sensible style checks as optional CI tasks,
- removes Travis CI, and
- applies a few code style fixes.

The new CI job should be optional by default, because I think one has to mark them manually as required for PRs.

PR #95, opened 2021-10-02
multiboot2-header crate - initial release

phip1611:
After I prepared everything, the `multiboot2-header` crate is ready for the initial release. I've put a lot of work into it and I am confident that the code base is solid. There are some tests and I also used external tools to compare the results, how example how `bootinfo` parses the mb2 header.

Any additional comments @IsaacWoods ?


multiboot2/src/lib.rs
```diff
@@ -389,7 +389,7 @@ impl Reader {
 
     pub(crate) fn read_u8(&mut self) -> u8 {
         self.off += 1;
-        unsafe { core::ptr::read(self.ptr.add(self.off - 1)) }
```
phip1611, line 392: oops,  this was committed by accident. But this shouldn't change everything, because dereferencing a u8 or reading it, result both in a copy.

PR #97, opened 2022-03-18
Updated 12.0 changelog to note breaking change to iterators

drzewiec:
I noticed that in PR #72 a change was made which is a breaking change, but was not mentioned as such in the changelog. Updated changelog to reflect this.


phip1611:
Thanks, good catch! @drzewiec

drzewiec:
Can do. I couldn't find a tag for 11.0 so I wasn't able to tell whether this change was in that version. Will push up an update shortly.

phip1611 (approved):

Issue #98, opened 2022-05-02
multiboot2-header is not `no_std`


chbaker0:
The docs claim the library is `no_std` in the second sentence of the front page: https://github.com/rust-osdev/multiboot2/blob/main/multiboot2-header/README.md?plain=1#L7

However the code references `std` in multiple places:

`Multiboot2HeaderBuilder::build()`: https://github.com/rust-osdev/multiboot2/blob/main/multiboot2-header/src/header/builder.rs#L129

`std` is `use`d in the beginning of lib.rs: https://github.com/rust-osdev/multiboot2/blob/main/multiboot2-header/src/lib.rs#L74

When I got build errors using this I was hoping there would be a `std` feature I could turn off, but it doesn't exist :(

The docs should probably be updated

phip1611:
Ohhh... Thank you for the hint! I'll fix it asap so that it is no_std and include a test in CI

phip1611:
@chbaker0 fixed in v0.1.1! public on crates.io

chbaker0:
Awesome! Thanks for the quick response!

PR #101, opened 2022-05-02
repo: add CI no_std check

phip1611:
Currently, the CI does not check if this actually builds with `no_std`. This MR fixes this.

After #98 is fixed, this should be mandatory for all new MRs


phip1611:
@IsaacWoods @phil-opp  I do not have admin rights. Can you please add the `build_no_std`-CI step to mandatory checks for MRs?

This PR arises from #98 . With this change, we can verify that the crate builds in `no_std` contexts.

PR #106, opened 2022-05-02
publicly announce the MSRV is 1.52.1.

phip1611:
I think this is relevant information that should be announced officially. This also helps us to identify breaking changes when we introduce new Rust functionality.


multiboot2-header/src/information_request.rs
```diff
@@ -160,7 +160,7 @@ impl InformationRequestHeaderTagBuilder {
 
         for tag in &self.irs {
             let bytes: [u8; 4] = (*tag as u32).to_ne_bytes();
-            data.extend(bytes);
+            data.extend(&bytes);
```
phip1611, line 163: without this, MSRV is 1.53.0 (but only for the header-crate)

PR #119, opened 2023-02-24
multiboot2: Fix EFI Memory Map 'last_area' calculation

A0lson:
Correct EFI 'last_area' calculation and align it to be more similar the E820 version.

(Previously, iterating the EFI Memory map resulted in a superfluous entry as it ran over the next tag)

### DIFF
--- a/multiboot2/src/memory_map.rs
+++ b/multiboot2/src/memory_map.rs
@@ -18,7 +18,7 @@ pub struct MemoryMapTag {
     size: u32,
     entry_size: u32,
     entry_version: u32,
-    first_area: MemoryArea,
+    first_area: [MemoryArea; 0],
 }
 
 impl MemoryMapTag {
@@ -31,10 +31,13 @@ impl MemoryMapTag {
     /// Return an iterator over all marked memory areas.
     pub fn all_memory_areas(&self) -> impl Iterator<Item = &MemoryArea> {
         let self_ptr = self as *const MemoryMapTag;
-        let start_area = (&self.first_area) as *const MemoryArea;
+        let start_area = self.first_area.as_ptr();
+
         MemoryAreaIter {
             current_area: start_area as u64,
-            last_area: (self_ptr as u64 + (self.size - self.entry_size) as u64),
+            // NOTE: `last_area` is only a bound, it doesn't necessarily point exactly to the last element
+            last_area: (self_ptr as u64
+                + (self.size as u64 - core::mem::size_of::<MemoryMapTag>() as u64)),
             entry_size: self.entry_size,
             phantom: PhantomData,
         }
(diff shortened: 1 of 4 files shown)


multiboot2/src/memory_map.rs
```diff
@@ -140,7 +140,7 @@ impl EFIMemoryMapTag {
         let start_area = (&self.first_desc) as *const EFIMemoryDesc;
         EFIMemoryAreaIter {
             current_area: start_area as u64,
-            last_area: (self_ptr as u64 + self.size as u64),
+            last_area: (self_ptr as u64 + (self.size - self.desc_size) as u64),
```
phip1611, line 143: I think that the root problem is that `size_of::<EFIMemoryMapTag>()` contains already the size of the first entry. I think a nicer solution would be to type the `EFIMemoryMapTag` as

```rust
pub struct EFIMemoryMapTag {
    typ: TagType,
    size: u32,
    desc_size: u32,
    desc_version: u32,
    first_desc: [EFIMemoryDesc; 0],
}
```

instead. Depending on your skill and time you'd like to invest, feel free to look into it. Otherwise, I can take care of it.

phip1611:
Hi and thanks a lot for your contribution! There are unit-test failures, but it might also be the case, that the unit test itself is wrong. I can't look into it right now.

If you'd like to invest more time into it, I'd appreciate it. Otherwise, I try to look into it within the next 10 days.

multiboot2/src/memory_map.rs
A0lson: Thanks for the suggestion. I'm a bit new to Rust, but the zero length arrays seem much more elegant.

I've updated both the E820 and EFI code to use them.

multiboot2/src/memory_map.rs
```diff
@@ -137,10 +140,12 @@ impl EFIMemoryMapTag {
     /// available memory areas for tables and such.
     pub fn memory_areas(&self) -> EFIMemoryAreaIter {
         let self_ptr = self as *const EFIMemoryMapTag;
-        let start_area = (&self.first_desc) as *const EFIMemoryDesc;
+        let start_area = self.first_desc.as_ptr();
         EFIMemoryAreaIter {
             current_area: start_area as u64,
-            last_area: (self_ptr as u64 + self.size as u64),
+            // NOTE: `last_area` is only a bound, it doesn't necessarily point exactly to the last element
```
phip1611, line 146: Why not? 🤔

multiboot2/src/memory_map.rs
phip1611: looks good!

phip1611:
thanks! the bug fix is live on crates.io!

multiboot2/src/memory_map.rs
A0lson: An oddly formed multiboot tag could have a combination of "tag size" and "EFI descriptor size" that represents a non-integer multiple of descriptors.  Thus, I wanted to make it clear that "last_area" should only be used as a bound and not a direct pointer to the last descriptor.

PR #128, opened 2023-03-18
ci: run miri + adjustments for miri

### LINKED ISSUE #117: CI: Run Tests with Miri to Catch Memory Issues
We should add `cargo miri test` to the CI. When I run it locally, miri complains about memory safety issues.

phip1611:
Closes #117

### DIFF
--- a/.github/workflows/_build-rust.yml
+++ b/.github/workflows/_build-rust.yml
@@ -50,6 +50,11 @@ on:
         required: false
         default: true
         description: Execute tests.
+      do-miri:
+        type: boolean
+        required: false
+        default: false
+        description: Execute unit tests with miri.
 
 jobs:
   rust:
@@ -104,3 +109,10 @@ jobs:
           Expand-Archive .\cargo-nextest.zip
           cp .\cargo-nextest/cargo-nextest.exe .
           .\cargo-nextest.exe nextest run --features ${{ inputs.features }} --no-default-features
+      - name: Unit Test with Miri
+        if: inputs.do-miri
+        # "--tests" so that the doctests are skipped. Currently, the doctest
+        # in miri fails.
+        run: |
+          rustup component add miri
+          cargo miri test --tests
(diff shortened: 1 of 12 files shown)


phip1611:
Could I ask you @nicholasbishop please for help? You do have more experience with miri than I have, I think. I don't quite understand why miri fails and if this is valid or how I should either fix or it. Do you have some advice? 

You can find a failed run here: <https://github.com/rust-osdev/multiboot2/actions/runs/4526183512/jobs/7971182214>

nicholasbishop:
I think the error in `multiboot2/src/boot_loader_name.rs:34` is due to https://github.com/rust-lang/unsafe-code-guidelines/issues/256. I haven't read through that full thread, so there are probably more ways of handling this, but what I did in uefi-rs is convert the header type to a DST. So the `BootLoaderNameTag.string` field would become a `[u8]`. See https://github.com/rust-osdev/uefi-rs/blob/main/uefi/src/proto/device_path/mod.rs#L126 for one example of this.

phip1611:
At this state, miri is somewhat useful. All tests that use the `cast_tag()` function (transitively) or the `TagIter`, need the `#[cfg_attr(miri, ignore)]` attribute.. I think that we are memory safe and miri just can't proof it.

A possible solution might be to model the whole parsing logic differently:
1. get pointer to mbi
2. find end
3. create a slice
4. operate on that slice. Then, all mem options are part of the "borrowed memory stack", according to miri

phip1611:
This MR now introduces basic miri support. Most tests are skipped by miri. There are two major issues:

- Each test that uses the `cast_tag` (transitively) fails, as miri can't guarantee that the memory is valid, unfortunately
- when something like `struct_as_bytes() ... cast` is done, there are alignment issues as the returned Vec<u8> is of course only aligned to a one byte boundary... so far, in tests this never was a problem, luckily. At least on x86.

PR #129, opened 2023-03-18
Add Integration Test

phip1611:
This adds integration tests for the multiboot2 crate and the multiboot2-header crate.


phip1611:
I'm very thankful for your contribution, @YtvwlD - now, I can write a much better integration test by constructing a real MBI :)

YtvwlD:
Oh cool. The chainloader is an interesting idea.

Issue #132, opened 2023-03-27
Switch to DSTs


phip1611:
After this was discussed in https://github.com/rust-osdev/multiboot2/pull/118#issuecomment-1330647495 and https://github.com/rust-osdev/multiboot2/pull/128#issuecomment-1484208223, I came to the conclusion, that all such structs should be refactored to DSTs. This will satisfy miri and also cleans up code.

phip1611:
The feature in Rust is unstable, but luckily, there is <https://crates.io/crates/ptr_meta>

YtvwlD:
oh, I've done this in #133 :)

phip1611:
Oh, I did not see this at a first glance. Now we also have #134  😬

PR #133, opened 2023-04-17
Add a builder to multiboot2

YtvwlD:
Hey, I'm currently working on adding Multiboot 2 support to a bootloader ([towboot](https://github.com/hhuOS/towboot)) and while using the multiboot2 and multiboot2-header crates, I encountered some pieces that were missing.

I particular:
 * I added more getters to multiboot2-header (mostly inspired by the ones in multiboot2).
 * I added a builder to multiboot2 (inspired by multiboot2-header).
 * I added the SMBIOS tag.
 * I added a function to find a header in a given (file) slice.

This is mostly just boilerplate, but there were two bigger problems I had to work around:

1) The memory map is only really accurate after exiting Boot Services, but then the bootloader can't allocate anymore.

I've solved this by prefilling the memory map tags with empty entries and overwriting them later on. It seems to work, but it's not pretty.

2) Most tags are dynamically sized. The previous implementation had them fixed-size in Rust and let the content dangle behind them. I think this is a bit weird and it doesn't work when you try to allocate new structs.

I made them dynamically sized – which is worse to handle in Rust (you can't put them on the stack anymore).
In addition, since the `ptr_metadata` feature is currently unstable, I used the slice methods instead. (You can probably revert 800567f8974ab430a012eb01918f716f762bf8e9 once it's stable).

I tried to adjust the tests and add some where appropriate. If that's not enough, I'll add more. :)

There's one test that fails for me on stable Rust (rustc 1.68.0 (2c8cc3432 2023-03-06)), though it works on nightly (rustc 1.70.0-nightly (900c35403 2023-03-08)).

I tried to keep the changes small. I think eg. that it would be nicer if `Multiboot2Header::from_addr` returned a `Result`, but I didn't change it

### DIFF
--- a/multiboot2/Cargo.toml
+++ b/multiboot2/Cargo.toml
@@ -32,7 +32,10 @@ repository = "https://github.com/rust-osdev/multiboot2"
 documentation = "https://docs.rs/multiboot2"
 
 [features]
-default = []
+# by default, builder is included
+default = ["builder"]
+std = []
+builder = ["std"]
 # Nightly-only features that will eventually be stabilized.
 unstable = []
 
--- a/multiboot2/src/boot_loader_name.rs
+++ b/multiboot2/src/boot_loader_name.rs
@@ -4,7 +4,10 @@ use core::mem::size_of;
 use core::str::Utf8Error;
 
 #[cfg(feature = "builder")]
-use {crate::builder::boxed_dst_tag, alloc::boxed::Box, alloc::vec::Vec};
+use {
+    crate::builder::boxed_dst_tag, crate::builder::traits::StructAsBytes, alloc::boxed::Box,
+    alloc::vec::Vec,
+};
 
 const METADATA_SIZE: usize = size_of::<TagTypeId>() + size_of::<u32>();
 
@@ -63,6 +66,13 @@ impl TagTrait for BootLoaderNameTag {
     }
 }
 
+#[cfg(feature = "builder")]
+impl StructAsBytes for BootLoaderNameTag {
+    fn byte_size(&self) -> usize {
+        self.size.try_into().unwrap()
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use crate::{BootLoaderNameTag, Tag, TagType};
@@ -94,5 +104,16 @@ mod tests {
         let tag = tag.cast_tag::<BootLoaderNameTag>();
         assert_eq!({ tag.typ }, TagType::BootLoaderName);
         assert_eq!(tag.name().expect("must be valid UTF-8"), MSG);
+    }
+
+    /// Test to generate a tag from a given string.
+    #[test]
+    #[cfg(feature = "builder")]
+    fn test_build_str() {
+        use crate::builder::traits::StructAsBytes;
+
+        let tag = BootLoaderNameTag::new(MSG);
+        let bytes = tag.struct_as_bytes();
+        assert_eq!(bytes, get_bytes());
     }
 }
--- a/multiboot2/src/boot_loader_name.rs
+++ b/multiboot2/src/boot_loader_name.rs
@@ -1,7 +1,12 @@
-use crate::TagTrait;
-use crate::{Tag, TagTypeId};
+use crate::{Tag, TagTrait, TagType, TagTypeId};
 use core::fmt::{Debug, Formatter};
+use core::mem::size_of;
 use core::str::Utf8Error;
+
+#[cfg(feature = "builder")]
+use {crate::builder::boxed_dst_tag, alloc::boxed::Box, alloc::vec::Vec};
+
+const METADATA_SIZE: usize = size_of::<TagTypeId>() + size_of::<u32>();
 
 /// The bootloader name tag.
 #[derive(ptr_meta::Pointee)]
@@ -14,6 +19,13 @@ pub struct BootLoaderNameTag {
 }
 
 impl BootLoaderNameTag {
+    #[cfg(feature = "builder")]
+    pub fn new(name: &str) -> Box<Self> {
+        let mut bytes: Vec<_> = name.bytes().collect();
+        bytes.push(0);
+        boxed_dst_tag(TagType::BootLoaderName, &bytes)
+    }
+
     /// Reads the name of the bootloader that is booting the kernel as Rust
     /// string slice without the null-byte.
     ///
@@ -46,10 +58,8 @@ impl Debug for BootLoaderNameTag {
 
 impl TagTrait for BootLoaderNameTag {
     fn dst_size(base_tag: &Tag) -> usize {
-        // The size of the sized portion of the bootloader name tag.
-        let tag_base_size = 8;
-        assert!(base_tag.size >= 8);
-        base_tag.size as usize - tag_base_size
+        assert!(base_tag.size as usize >= METADATA_SIZE);
+        base_tag.size as usize - METADATA_SIZE
     }
 }
 
--- a/multiboot2/src/builder/information.rs
+++ b/multiboot2/src/builder/information.rs
@@ -2,7 +2,7 @@
 use crate::builder::traits::StructAsBytes;
 use crate::{
     BasicMemoryInfoTag, BootLoaderNameTag, CommandLineTag, ElfSectionsTag, FramebufferTag,
-    ModuleTag,
+    MemoryMapTag, ModuleTag,
 };
 
 use alloc::boxed::Box;
@@ -18,6 +18,7 @@ pub struct Multiboot2InformationBuilder {
     command_line_tag: Option<Box<CommandLineTag>>,
     elf_sections_tag: Option<Box<ElfSectionsTag>>,
     framebuffer_tag: Option<Box<FramebufferTag>>,
+    memory_map_tag: Option<Box<MemoryMapTag>>,
     module_tags: Vec<Box<ModuleTag>>,
 }
 
@@ -29,6 +30,7 @@ impl Multiboot2InformationBuilder {
             command_line_tag: None,
             elf_sections_tag: None,
             framebuffer_tag: None,
+            memory_map_tag: None,
             module_tags: Vec::new(),
         }
     }
@@ -51,6 +53,10 @@ impl Multiboot2InformationBuilder {
 
     pub fn framebuffer_tag(&mut self, framebuffer_tag: Box<FramebufferTag>) {
         self.framebuffer_tag = Some(framebuffer_tag);
+    }
+
+    pub fn memory_map_tag(&mut self, memory_map_tag: Box<MemoryMapTag>) {
+        self.memory_map_tag = Some(memory_map_tag);
     }
 
     pub fn add_module_tag(&mut self, module_tag: Box<ModuleTag>) {
--- a/multiboot2/src/builder/information.rs
+++ b/multiboot2/src/builder/information.rs
@@ -1,8 +1,8 @@
 //! Exports item [`Multiboot2InformationBuilder`].
 use crate::builder::traits::StructAsBytes;
 use crate::{
-    BasicMemoryInfoTag, BootInformationInner, BootLoaderNameTag, CommandLineTag, ElfSectionsTag,
-    EndTag, FramebufferTag, MemoryMapTag, ModuleTag,
+    BasicMemoryInfoTag, BootInformationInner, BootLoaderNameTag, CommandLineTag, EFISdt32,
+    EFISdt64, ElfSectionsTag, EndTag, FramebufferTag, MemoryMapTag, ModuleTag,
 };
 
 use alloc::boxed::Box;
@@ -21,6 +21,8 @@ pub struct Multiboot2InformationBuilder {
     framebuffer_tag: Option<Box<FramebufferTag>>,
     memory_map_tag: Option<Box<MemoryMapTag>>,
     module_tags: Vec<Box<ModuleTag>>,
+    efisdt32: Option<EFISdt32>,
+    efisdt64: Option<EFISdt64>,
 }
 
 impl Multiboot2InformationBuilder {
@@ -29,6 +31,8 @@ impl Multiboot2InformationBuilder {
             basic_memory_info_tag: None,
             boot_loader_name_tag: None,
             command_line_tag: None,
+            efisdt32: None,
+            efisdt64: None,
             elf_sections_tag: None,
             framebuffer_tag: None,
             memory_map_tag: None,
@@ -64,6 +68,12 @@ impl Multiboot2InformationBuilder {
             len += Self::size_or_up_aligned(tag.byte_size())
         }
         if let Some(tag) = &self.command_line_tag {
+            len += Self::size_or_up_aligned(tag.byte_size())
+        }
+        if let Some(tag) = &self.efisdt32 {
+            len += Self::size_or_up_aligned(tag.byte_size())
+        }
+        if let Some(tag) = &self.efisdt64 {
             len += Self::size_or_up_aligned(tag.byte_size())
         }
         if let Some(tag) = &self.elf_sections_tag {
@@ -117,6 +127,12 @@ impl Multiboot2InformationBuilder {
         if let Some(tag) = self.command_line_tag.as_ref() {
             Self::build_add_bytes(&mut data, &tag.struct_as_bytes(), false)
         }
+        if let Some(tag) = self.efisdt32.as_ref() {
+            Self::build_add_bytes(&mut data, &tag.struct_as_bytes(), false)
+        }
+        if let Some(tag) = self.efisdt64.as_ref() {
+            Self::build_add_bytes(&mut data, &tag.struct_as_bytes(), false)
+        }
         if let Some(tag) = self.elf_sections_tag.as_ref() {
             Self::build_add_bytes(&mut data, &tag.struct_as_bytes(), false)
         }
@@ -145,6 +161,14 @@ impl Multiboot2InformationBuilder {
 
     pub fn command_line_tag(&mut self, command_line_tag: Box<CommandLineTag>) {
         self.command_line_tag = Some(command_line_tag);
+    }
+
+    pub fn efisdt32(&mut self, efisdt32: EFISdt32) {
+        self.efisdt32 = Some(efisdt32);
+    }
+
+    pub fn efisdt64(&mut self, efisdt64: EFISdt64) {
+        self.efisdt64 = Some(efisdt64);
     }
 
     pub fn elf_sections_tag(&mut self, elf_sections_tag: Box<ElfSectionsTag>) {
--- a/multiboot2/src/builder/information.rs
+++ b/multiboot2/src/builder/information.rs
@@ -2,7 +2,8 @@
 use crate::builder::traits::StructAsBytes;
 use crate::{
     BasicMemoryInfoTag, BootInformationInner, BootLoaderNameTag, CommandLineTag, EFISdt32,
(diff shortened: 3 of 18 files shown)


phip1611:
Hi @YtvwlD 
Thanks very much for working on this. I'm in big favor of the new functionality. Can you please rebase your changes onto the latest main branch? I refrain from merging any other breaking stuff, until we get this thing merged - to keep things simple.

I will have a more detailed look in the upcoming days.

YtvwlD:
That's great to hear. :)

I've tried rebasing this on Monday, but it didn't compile afterwards, sadly. I hope to be able to do it this weekend.

YtvwlD:
So, this took a bit longer than expected. :)
The one test sadly still only works with some compilers (though it seems to work on stable and fail on nightly in the CI, interesting).

I've removed `cast_tag` after the rebase because rustc doesn't seem to support generic functions over dynamically sized types? Here's what I was getting:

```rust
error[E0606]: casting `*const Tag` as `*const T` is invalid
   --> multiboot2/src/tag_type.rs:307:20
    |
307 |         unsafe { &*(self as *const Tag as *const T) }
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: vtable kinds may not match
```

(I've tried to put a [minimal example in the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=70bcaadd80c7d2cd067d18753e3443db).)

phip1611:
> I've removed cast_tag after the rebase because rustc doesn't seem to support generic casts over dynamically sized types?

Yes, because DSTs uses fat pointers. This is also an existing problem in the multiboot2 crate. I plan to refactor this eventually - If I find some time.. One can fix this with https://crates.io/crates/ptr_meta for example.

YtvwlD:
`cast_tag` didn't exist before the rebase. Each getters had it's own cast.

phip1611:
Ah yes, totally right. I broke the API with that change... the right solution would be to remove cast_tag() and properly support DSTs... a quick fix for you might be to add something like `find_tag` next to `get_tag` that has the right semantics. This way, you could make your code compile.

YtvwlD:
ah, I've just reverted the two commits that added it (the last two commits in this PR).

YtvwlD:
The msrv fail is because `alloc::ffi` is only available since [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#c-compatible-ffi-types-in-core-and-alloc). Any ideas on what to do about that?

phip1611:
I started working on DSTs which should fix your 

```
error[E0606]: casting `*const Tag` as `*const T` is invalid
   --> multiboot2/src/tag_type.rs:307:20
    |
307 |         unsafe { &*(self as *const Tag as *const T) }
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: vtable kinds may not match
````

problem. What do you think about my changes in #132 ? We could also merge your PR first and I adapt it then to my changes. But I'd like to hear if the new API would work in your case. I still hadn't much time to look into your code, unfortunately 

> The msrv fail is because alloc::ffi is only available since [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#c-compatible-ffi-types-in-core-and-alloc). Any ideas on what to do about that?

I don't have anything against a MSVR bump, but I do not see a benefit in using this abstraction here. Multiboot2 strings and Rust strings are guaranteed to be valid UTF-8, CStrings are not. So I think it would be better to use regular Rust strings and push a terminating `\0` onto the rust string. When the data is serialized (with `.as_bytes()`), the terminating null will be included.

**Comment by YtvwlD** [comment]
I don't have any problems with fat pointers (anymore), because here, `Tag` is unsized, too. Casting  `Tag` to eg. `BootLoaderNameTag` works out of the box. (Unless that cast is generic, then the `vtable` compiler error occurs.)

Casting raw bytes to a struct still requires explicit fat pointers, though (as can be seen in the tests for `BootLoaderNameTag`).

Having a `from_base_tag` method in a trait might be better than me just removing `cast_tag`, though.

I'd rebase this PR to keep using `[u8]` for strings then, but it'll probably make more sense for you to look into it, first?

File: multiboot2-header/src/header.rs [hunk 1]
```diff
@@ -59,6 +60,28 @@ impl<'a> Multiboot2Header<'a> {
         Self { inner: reference }
     }
 
+    /// Find the header in a given slice.
+    pub fn find_header(buffer: &[u8]) -> Option<(&[u8], u32)> {
+        // the magic is 32 bit aligned and inside the first 8192 bytes
```
phip1611, line 66: There should be an assertion for correct alignment

multiboot2-header/src/header.rs
```diff
@@ -59,6 +60,28 @@ impl<'a> Multiboot2Header<'a> {
...
+        let mut chunks = buffer[0..8192].chunks_exact(4);
+        let magic_index = match chunks.position(|vals| {
+            u32::from_le_bytes(vals.try_into().unwrap()) // yes, there's 4 bytes here
+            == MULTIBOOT2_HEADER_MAGIC
+        }) {
+            Some(idx) => idx * 4,
+            None => return None,
+        };
+        chunks.next(); // arch
+        let header_length: usize = u32::from_le_bytes(chunks.next().unwrap().try_into().unwrap())
```
phip1611, line 77: A lot of unwraps in one line. I think, this could could be written a little nicer. Perhaps, returning a Result would be better?

multiboot2/Cargo.toml
```diff
@@ -32,7 +32,10 @@ repository = "https://github.com/rust-osdev/multiboot2"
 documentation = "https://docs.rs/multiboot2"
 
 [features]
-default = []
+# by default, builder is included
+default = ["builder"]
+std = []
```
phip1611, line 37: I think, but I'm not sure right now, that we do not need a std module but only an alloc module. Which code from the standard library do you need for the builder which doesn't work with no_std+alloc?

phip1611:
I'm on vacation right now, so I won't have much capacity for more review in the next 7 days.

YtvwlD:
Have fun. :)

multiboot2/Cargo.toml
phip1611: Yeah, please name the module `alloc`. I'll fix this in the header crate as well. It makes no sense to depend on `std`

phip1611: You can leave it as it is. I'll refactor this in #135

multiboot2/src/command_line.rs
```diff
@@ -1,24 +1,48 @@
...
+#[cfg(feature = "builder")]
+use crate::builder::boxed_dst_tag;
+use crate::tag_type::{TagType, TagTypeId};
+use core::convert::TryInto;
 use core::mem;
 use core::slice;
 use core::str;
 
+#[cfg(feature = "builder")]
+use alloc::ffi::CString;
```
phip1611, line 16: please use core::alloc::String and in the end, manually push `\0`.

multiboot2/src/elf_sections.rs
```diff
@@ -10,30 +15,31 @@ const METADATA_SIZE: usize = mem::size_of::<TagTypeId>() + mem::size_of::<u32>()
 ///
 /// The sections iterator is provided via the `sections` method.
 #[derive(Debug)]
+#[repr(C, packed)]
 pub struct ElfSectionsTag {
```
phip1611, line 17: interesting that the ElfSectionsTag did have an other layout so far.

multiboot2/src/lib.rs
```diff
@@ -1305,12 +1301,12 @@ mod tests {
...
-        assert_eq!(fbi.width, 80);
-        assert_eq!(fbi.height, 25);
-        assert_eq!(fbi.bpp, 16);
-        assert_eq!(fbi.buffer_type, FramebufferType::Text);
+        assert_eq!(fbi.address(), 753664);
+        assert_eq!(fbi.pitch(), 160);
+        assert_eq!(fbi.width(), 80);
+        assert_eq!(fbi.height(), 25);
+        assert_eq!(fbi.bpp(), 16);
+        assert_eq!(fbi.buffer_type().unwrap(), FramebufferType::Text);
```
phip1611, line 1375: nit: Matching against `Some(FramebufferType::Text)` is better code style IMHO

multiboot2/src/builder/information.rs
```diff
@@ -63,3 +163,41 @@ impl Multiboot2InformationBuilder {
...
+        builder.add_module_tag(ModuleTag::new(0, 1234, "module1"));
+        builder.add_module_tag(ModuleTag::new(5678, 6789, "module2"));
+
+        println!("builder: {:#?}", builder);
+        println!("expected_len: {} bytes", builder.expected_len());
+
+        let mb2i_data = builder.build();
+        let mb2i_addr = mb2i_data.as_ptr() as usize;
+        let mb2i = unsafe { load(mb2i_addr) };
+        println!("{:#?}", mb2i);
```
phip1611, line 323: nit: maybe, a more sophisticated test is what we want here

multiboot2/src/builder/traits.rs
```diff
@@ -1,16 +1,13 @@
...
-use core::mem::size_of;
-
 /// Trait for all tags that helps to create a byte array from the tag.
 /// Useful in builders to construct a byte vector that
 /// represents the Multiboot2 information with all its tags.
-pub(crate) trait StructAsBytes: Sized {
-    /// Returns the size in bytes of the struct, as known during compile
-    /// time. This doesn't use read the "size" field of tags.
-    fn byte_size(&self) -> usize {
-        size_of::<Self>()
```
phip1611: why did you do this change? This is unclear to me

multiboot2/src/boot_loader_name.rs
```diff
@@ -20,7 +21,7 @@ const METADATA_SIZE: usize = mem::size_of::<TagTypeId>() + mem::size_of::<u32>()
 ///
 /// The name is a normal C-style UTF-8 zero-terminated string that can be
 /// obtained via the `name` method.
-#[derive(Debug)]
+
```
phip1611, line 24: empty line

multiboot2/src/lib.rs
```diff
@@ -262,6 +262,12 @@ impl BootInformation {
         self.get_tag::<MemoryMapTag, _>(TagType::Mmap)
     }
 
+    /// Search for the Memory map tag, return a mutable reference.
+    pub fn memory_map_tag_mut(&mut self) -> Option<&mut MemoryMapTag> {
```
phip1611, line 271: yikes, this one is controversial for me. Why do you want to change the multiboot2 boot information? I do not like the idea.. I think, a more clean approach would be to copy an owned version into your kernel, bootloader or whatever and modify the data there.

phip1611:
I've had time to review everything. I converted this MR as draft, as there are several points to address before getting this merged. I hope I do not make the contribution process for you to frustrating, but this PR is in its current state to big to pass the review process.

I'd prefer to have dedicated PRs for the following things:
  - add missing functionality in multiboot2-header
    - find the header
    - add getter for the tags
    - that module align thingy 
  - add missing tags in multiboot2 
    - add smbios tag
    - add basic memory info tag
  - add multiboot2 MBI builder
  - mutable references to the mbi (**this one is controversial to me, so a dedicated MR where we can discuss**

I think, and as you already said, that the DST approach I took in #132 is more mature. Maybe you can have a look there and tell me if it would work with your contribution. Then we merge it and you can rebase your code onto those changes.

I'd highly appreciate if you can address (at least some of) my concerns and improvement suggestions. Otherwise, I'd assist you and do (some) of the work myself.

Don't get me wrong, U'm are very happy of merging the changes. But this PR is just to big.

YtvwlD:
Okay, but I'll probably need a bit of time.

I hope to be able to work on the multiboot2-header PR on the next weekend (as this should be pretty easy and uncontroversial).

The other stuff probably depends on DSTs, so it would be better to discuss that.

(Wir können da auch gerne per Video drüber quatschen, aber ich kann leider entweder heute am späten Nachmittag / Abend oder erst wieder Montag.)

YtvwlD:
(This is just a rebase, nothing to see here. :)

multiboot2/src/builder/information.rs
YtvwlD: yeah, probably? This is on the same level as the test in multiboot2-header (as I've copied it from there).

multiboot2/src/builder/traits.rs
YtvwlD: I'm not entirely sure. Is this still relevant?

multiboot2/src/lib.rs
YtvwlD: I'm splitting this off for now. :)

YtvwlD:
So, that's another rebase. :)

I've split the mutable stuff off into three new commits, let's talk about them later.

multiboot2/src/builder/mod.rs
```diff
@@ -4,3 +4,41 @@ mod information;
...
+use alloc::boxed::Box;
+use core::alloc::Layout;
+use core::mem::size_of;
+
+use crate::{TagTrait, TagTypeId};
+
+/// Create a boxed tag with the given content.
+pub(super) fn boxed_dst_tag<T>(typ: TagTypeId, content: &[u8]) -> Box<T>
+where
+    T: TagTrait + ?Sized + ptr_meta::Pointee<Metadata = usize>,
```
phip1611, line 18: nit: I think you can also do sth like `TagTrait<Metadata = usize>` as `TagTrait` depends on `Pointee`.

multiboot2/src/builder/mod.rs
```diff
@@ -4,3 +4,41 @@ mod information;
...
+
+use alloc::alloc::alloc;
+use alloc::boxed::Box;
+use core::alloc::Layout;
+use core::mem::size_of;
+
+use crate::{TagTrait, TagTypeId};
+
+/// Create a boxed tag with the given content.
+pub(super) fn boxed_dst_tag<T>(typ: TagTypeId, content: &[u8]) -> Box<T>
```
phip1611, line 16: `typ: TagTypeId` could be `Into<TagTypeId>`

multiboot2/src/command_line.rs
```diff
@@ -18,6 +28,15 @@ pub struct CommandLineTag {
 }
 
 impl CommandLineTag {
+    /// Create a new command line tag from the given string.
+    #[cfg(feature = "builder")]
+    pub fn new(command_line: &str) -> Box<Self> {
+        let mut bytes: Vec<_> = command_line.bytes().collect();
+        bytes.push(0);
+        let tag = boxed_dst_tag(TagType::Cmdline.into(), &bytes);
+        unsafe { Box::from_raw(Box::into_raw(tag) as *mut Self) }
```
phip1611, line 41: `unsafe { Box::from_raw(Box::into_raw(tag) as *mut Self) }`

I think this should be abstracted away in the `boxed_dst_tag` crate, right? Or is there a technical limitation?

multiboot2/src/boot_loader_name.rs
```diff
@@ -1,8 +1,19 @@
-use crate::TagTrait;
-use crate::{Tag, TagTypeId};
+use crate::{Tag, TagTrait, TagType, TagTypeId};
 use core::fmt::{Debug, Formatter};
+use core::mem::size_of;
 use core::str::Utf8Error;
 
+#[cfg(feature = "builder")]
```
phip1611, line 6: could be

```rust
#[cfg(feature = "builder")]
use {
    crate::builder::boxed_dst_tag,
    use alloc::vec::Vec,
    alloc::boxed::Box,
}
```

multiboot2/src/memory_map.rs
```diff
@@ -13,17 +23,29 @@ use core::mem;
...
-#[derive(Debug)]
+#[derive(Debug, ptr_meta::Pointee)]
 #[repr(C)]
 pub struct MemoryMapTag {
     typ: TagTypeId,
     size: u32,
     entry_size: u32,
     entry_version: u32,
-    first_area: [MemoryArea; 0],
+    areas: [MemoryArea],
```
phip1611, line 34: ahh good catch! Thanks for fixing this

phip1611 (approved):
Awesome, thanks for working on this! Only a few nits left.

multiboot2/src/command_line.rs
YtvwlD: That was needed because of the [generic casts](https://github.com/rust-osdev/multiboot2/pull/133#issuecomment-1520421043), but it's not anymore with `TagTrait`, you're right!

multiboot2/src/builder/mod.rs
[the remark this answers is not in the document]
YtvwlD: You're right: Having tests for this really makes sense.

I've added unit tests for building a BootLoaderNameTag, CommandLineTag, ModuleTag and SmbiosTag (and catched some missing bytes in SmbiosTag that way).

I'm not sure that's enough? Do we want to test `boxed_dst_tag` by itself?

multiboot2/src/builder/information.rs
YtvwlD: I added more assertions here. :)

YtvwlD:
Do we want a `EFIMemoryDesc::new`? I'm currently just transmuting the ones from the firmware into it (well, and checking the size), but I'm not sure if that's the right way.

multiboot2/src/lib.rs
```diff
@@ -1262,7 +1265,7 @@ mod tests {
         assert_eq!(addr + bytes.len(), bi.end_address());
         assert_eq!(bytes.len(), bi.total_size());
         let es = bi.elf_sections_tag().unwrap();
-        let mut s = es.sections();
+        let mut s = es.sections(bi.offset);
```
phip1611, line 1282: It's not entirely clear why it is now required to have this parameter in the public API and which are the tyical values that users want to pass in.. @YtvwlD

phip1611: if possible, please add more documentation with some guidance

phip1611 (approved):
LGTM. Thanks! Please address that one single remark regarding the sections(n) method.

multiboot2/src/builder/mod.rs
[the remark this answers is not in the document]
phip1611: I think its fine

multiboot2/src/lib.rs
YtvwlD: Oh, yes, this is because `BootInformation::elf_sections_tag` returns the real sections tag since 41428fbe57c84958bdbc4cce810e35bc6ac71975 and not a wrapper, so I had to(?) remove the passing of `self.offset` there.

Perhaps this should be `BootInformation::elf_sections` and just return the iterator instead of the tag?
The tag has no other public fields or methods.

As for the offset: I don't really know. This is the same offset that's passed to `BootInformation::load_with_offset`, `0` would be a typical value.

YtvwlD: I've done this in the latest commit, let me know what you think.

phip1611 (approved):
Thanks for working on this. I'm glad we finally can merge it.

I address some open issues and then going to release it in a couple of days.

phip1611:
Via git bisecting, I could find that you might have broke something with commit https://github.com/rust-osdev/multiboot2/pull/133/commits/d198dceffe544b6324ba4c204270d9a1d934abba 

Since then, the debug print of the MBI doesn't work anymore and operates on wrong memory. I guess, some pointer magic is now wrong. I try to trace it down and create a unit test for it :)

PR #134, opened 2023-04-28
multiboot2: properly type DST tags

phip1611:
This PR introduces a proper DST typing of Multiboot2 tags that are DSTs. This includes all tags that have a dynamically size, usually used for a null-terminated string at the end of a tag.


multiboot2/src/boot_loader_name.rs
```diff
@@ -28,17 +29,31 @@ impl BootLoaderNameTag {
...
+            .field("name", &self.name())
+            .finish()
+    }
+}
+
+impl crate::TagTrait for BootLoaderNameTag {
+    fn dst_size(base_tag: &Tag) -> usize {
+        // The size of the sized portion of the bootloader name tag.
+        let tag_base_size = 8;
+        base_tag.size as usize - tag_base_size
```
phip1611, line 50: protect against invalid tag_base.size() values. panic if tag_base_size is bigger than tag_base_size

multiboot2/src/command_line.rs
```diff
@@ -33,16 +36,31 @@ impl CommandLineTag {
...
+            .field("cmdline", &self.command_line())
+            .finish()
+    }
+}
+
+impl TagTrait for CommandLineTag {
+    fn dst_size(base_tag: &Tag) -> usize {
+        // The size of the sized portion of the command line tag.
+        let tag_base_size = 8;
+        base_tag.size as usize - tag_base_size
```
phip1611, line 57: protect against invalid tag_base.size() values. panic if tag_base_size is bigger than tag_base_size

multiboot2/src/lib.rs
```diff
@@ -476,10 +493,49 @@ impl Reader {
...
+    /// sane and the underlying memory valid. The implementation of this trait
+    /// **must have** a correct [`Self::dst_size`] implementation.
+    unsafe fn from_base_tag<'a>(tag: &Tag) -> &'a Self {
+        let ptr = tag as *const _ as *const ();
+        let ptr = ptr_meta::from_raw_parts(ptr, Self::dst_size(tag));
+        &*ptr
+    }
+}
+
+// All sizes tags automatically have a Pointee implementation where
```
phip1611, line 525: nit: "all sized tags"

multiboot2/src/lib.rs
```diff
@@ -1490,46 +1538,54 @@ mod tests {
         consumer(MbiLoadError::IllegalAddress)
     }
 
+    /// Example for a custom tag without using a DST.
     #[test]
-    fn custom_tag() {
+    fn get_custom_tag_no_dst_from_mbi() {
```
phip1611, line 1543: this makes no sense. use a DST or drop the DST part.

multiboot2/src/module.rs
```diff
@@ -48,12 +44,21 @@ impl ModuleTag {
     }
 }
 
+impl crate::TagTrait for ModuleTag {
+    fn dst_size(base_tag: &Tag) -> usize {
+        // The size of the sized portion of the module tag.
+        let tag_base_size = 16;
+        base_tag.size as usize - tag_base_size
```
phip1611, line 51: protect against invalid tag_base.size() values. panic if tag_base_size is bigger than tag_base_size

multiboot2/src/lib.rs
```diff
@@ -476,10 +494,62 @@ impl Reader {
...
+    #[allow(clippy::unused_unit)]
+    fn dst_size(_: &Tag) -> Self::Metadata {
+        ()
+    }
+}
+
+// All sized tags automatically have a Pointee implementation where
+// Pointee::Metadata is (). Hence, the TagTrait is implemented automatically for
+// all tags that are sized.
+impl<T: Pointee<Metadata = usize>> TagTrait for T {
```
phip1611, line 539: hey @nicholasbishop,
May I ask you for help? That's a tough one, I think. I want to have default implementations for both cases but the compiler tells me the following

image

However, I'd assume that both implementations are distinct, as Metadata has a different type..

nicholasbishop: Possibly the advice in https://stackoverflow.com/questions/40392524 would help, although I had some difficulty following it.

I'm not quite clear on whether a blanket impl makes sense here anyway though. Do I have it right that all tag types must start with a 32-bit ID followed by a 32-bit size field? The blanket impl would mean that `TagTrait` is implemented for things like `()` and `u32` which seems wrong.

phip1611: I had a chat with a colleque. After looking at <https://github.com/rust-lang/rust/issues/20400> and <https://github.com/rust-lang/rfcs/pull/1672#issuecomment-1405377983>, I came to the conclusion that the solution without <https://github.com/rust-osdev/multiboot2/pull/134/commits/4a6964ed09ca6ad3caf36c28d9863ea7ad1bf556> is good enough.

phip1611:
@YtvwlD any further thoughts on this?

phip1611:
I had enough time to think about this, talk with people about it, and I think it is the best way we have at the moment.  

I'll merge it so the dependent MRs can make progress.

YtvwlD:
Sorry for responding that late, but yes, this looks fine to me. :)

PR #136, opened 2023-05-14
add missing functionality in multiboot2-header (finding the header, getting tags)

YtvwlD:
This is the first part split off of #133, as recommended by @phip1611 in https://github.com/rust-osdev/multiboot2/pull/133#issuecomment-1537145612:

> * find the header
> * add getter for the tags
> * that module align thingy

The getters are very inspired by the ones in `multiboot2`.

### DIFF
--- a/multiboot2-header/src/header.rs
+++ b/multiboot2-header/src/header.rs
@@ -397,6 +397,10 @@ impl Debug for Multiboot2HeaderTagIter {
                 let entry = t as *const EntryEfi64HeaderTag;
                 let entry = &*(entry);
                 debug.entry(entry);
+            } else if typ == HeaderTagType::ModuleAlign {
+                let entry = t as *const ModuleAlignHeaderTag;
+                let entry = &*(entry);
+                debug.entry(entry);
             } else if typ == HeaderTagType::Relocatable {
                 let entry = t as *const RelocatableHeaderTag;
                 let entry = &*(entry);
--- a/multiboot2-header/src/header.rs
+++ b/multiboot2-header/src/header.rs
@@ -1,7 +1,8 @@
 use crate::{
     AddressHeaderTag, ConsoleHeaderTag, EfiBootServiceHeaderTag, EndHeaderTag,
     EntryAddressHeaderTag, EntryEfi32HeaderTag, EntryEfi64HeaderTag, FramebufferHeaderTag,
-    HeaderTag, HeaderTagISA, HeaderTagType, InformationRequestHeaderTag, RelocatableHeaderTag,
+    HeaderTag, HeaderTagISA, HeaderTagType, InformationRequestHeaderTag, ModuleAlignHeaderTag,
+    RelocatableHeaderTag,
 };
 use core::convert::TryInto;
 use core::fmt::{Debug, Formatter};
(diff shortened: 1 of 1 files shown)


multiboot2-header/src/header.rs
```diff
@@ -59,6 +61,28 @@ impl<'a> Multiboot2Header<'a> {
         Self { inner: reference }
     }
 
+    /// Find the header in a given slice.
+    pub fn find_header(buffer: &[u8]) -> Option<(&[u8], u32)> {
```
YtvwlD, line 65: > There should be an assertion for correct alignment

I'm not exactly sure, how? The chunking skips 4 bytes each time, so unaligned headers just won't be found.

multiboot2-header/src/header.rs
```diff
@@ -59,6 +61,28 @@ impl<'a> Multiboot2Header<'a> {
...
+        let mut chunks = buffer[0..8192].chunks_exact(4);
+        let magic_index = match chunks.position(|vals| {
+            u32::from_le_bytes(vals.try_into().unwrap()) // yes, there's 4 bytes here
+            == MULTIBOOT2_HEADER_MAGIC
+        }) {
+            Some(idx) => idx * 4,
+            None => return None,
+        };
+        chunks.next(); // arch
+        let header_length: usize = u32::from_le_bytes(chunks.next().unwrap().try_into().unwrap())
```
YtvwlD, line 77: > A lot of unwraps in one line. I think, this could could be written a little nicer. Perhaps, returning a Result would be better?

The last unwrap shouldn't fail (it just converts 4 bytes to an `u32`), but the first one might, so yeah, returning an error for a header that is cut off might be sensible. What would be an appropriate error type here?

phip1611: I think an error similar to <https://github.com/rust-osdev/multiboot2/blob/main/multiboot2/src/lib.rs#L165> would be nice

phip1611: I could think of something like `-> Result<Option<Header>, BufError>`.

BufError could be something like `::ToSmall`, `::Unaligned` or so

multiboot2-header/src/header.rs
YtvwlD: Ah yes, it might be nicer to return an error instead of just `None` in the case of a misaligned header.

With that in mind: Wouldn't it also be nice if `Multiboot2Header::from_addr` returned a `Result` instead of panicking?

phip1611: ah yes, right. Probably it is the best option to add the `Result` type to <https://github.com/rust-osdev/multiboot2/blob/main/multiboot2-header/src/header.rs#L38> and then re-use that error type in the `find_header` function, aye?

I'm thinking of a new `multiboot2_header::LoadError` enum with the following variants:
- invalid address (must be 8-byte aligned and not null)
- MagicNotfound
- ChecksumMismatch

YtvwlD:
Hm, I think the parsing story is still a bit weird (pass the file, get the portion that is relevant as a slice, pin that and call `from_addr`), but it's workable and better than before. :)

I'd really like to just pass the file slice and get a reference to the header, but I don't know if this is workable with the inner references.

phip1611:
Please fix the minor CI problems and then we are good to go :)

phip1611:
Oh.

`#[allow(clippy::missing_const_for_fn)]` is fine - could you add it, please?

multiboot2-header/src/header.rs
```diff
@@ -97,6 +196,20 @@ impl<'a> Debug for Multiboot2Header<'a> {
     }
 }
 
+/// Errors that can occur when parsing a header from a slice.
+/// See [`Multiboot2Header::find_header`].
+#[derive(Debug)]
+pub enum LoadError {
```
phip1611, line 202: please also derive: `Clone, PartialEq, Eq`

phip1611:
> I didn't break [that](https://github.com/YtvwlD/multiboot2/actions/runs/4991371382/jobs/8937780563#step:9:35) failing lint. :)

yes, you're right. I'll fix that soon

phip1611 (approved):

PR #137, opened 2023-05-22
Add BasicMemoryInfo and SMBIOS tags

YtvwlD:
This is the first part split off of #133, as recommended by @phip1611 in https://github.com/rust-osdev/multiboot2/pull/133#issuecomment-1537145612:

> * add smbios tag
> * add basic memory info tag

I removed the `new` functions from the structs, this is just for reading.


phip1611:
Wonderful, thanks for working on this!

PR #140, opened 2023-06-19
Get mutable references to the memory information

YtvwlD:
This is (hopefully) the last part I split off of #133.

This PR adds the functions `basic_memory_info_tag_mut`, `memory_map_tag_mut` adn `efi_memory_map_tag_mut`. The memory map tags each get their own `memory_areas_mut`.

This may sound wrong at first and yes, this is really hacky, but I'm afraid that I need this.

The correct way to pass memory information from the bootloader to the kernel would be

1. gather memory information
2. build the appropriate tags
3. pass them into the builder
4. build the information struct

*But* steps 2 and 4 allocate memory which in turn might invalidate the gathered memory map.
One could either risk this and just pass possibly outdated information – or do it the other way round:

1. approximate how much memory areas there are
2. build the whole information struct
3. get the correct memory information
4. get the appropriate tags, mutably
5. put the information in there

(The downside to this approach is that there may be empty areas at the end (which seems to be fine) or that there are not enough areas.)

### DIFF
--- a/multiboot2/src/lib.rs
+++ b/multiboot2/src/lib.rs
@@ -99,7 +99,7 @@ use derive_more::Display;
 #[cfg(feature = "builder")]
 use crate::builder::AsBytes;
 use crate::framebuffer::UnknownFramebufferType;
-use tag::TagIter;
+use tag::{TagIter, TagIterMut};
 
 /// Magic number that a Multiboot2-compliant boot loader will use to identify
 /// the handoff. The location depends on the architecture and the targeted
@@ -152,9 +152,9 @@ impl AsBytes for BootInformationHeader {}
 
 /// This type holds the whole data of the MBI. This helps to better satisfy miri
 /// when it checks for memory issues.
-#[derive(ptr_meta::Pointee)]
+#[derive(ptr_meta::Pointee, Debug)]
 #[repr(C)]
-struct BootInformationInner {
+pub struct BootInformationInner {
     header: BootInformationHeader,
     tags: [u8],
 }
@@ -178,11 +178,22 @@ impl BootInformationInner {
     }
 }
 
+impl AsRef<BootInformationInner> for BootInformationInner {
+    fn as_ref(&self) -> &BootInformationInner {
+        self
+    }
+}
+
+impl AsMut<BootInformationInner> for BootInformationInner {
+    fn as_mut(&mut self) -> &mut BootInformationInner {
+        self
+    }
+}
 /// A Multiboot 2 Boot Information (MBI) accessor.
 #[repr(transparent)]
-pub struct BootInformation<'a>(&'a BootInformationInner);
+pub struct BootInformation<T: AsRef<BootInformationInner>>(T);
 
-impl<'a> BootInformation<'a> {
+impl BootInformation<&BootInformationInner> {
     /// Loads the [`BootInformation`] from a pointer. The pointer must be valid
     /// and aligned to an 8-byte boundary, as defined by the spec.
     ///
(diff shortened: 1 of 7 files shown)


multiboot2/src/lib.rs
```diff
@@ -249,6 +249,12 @@ impl BootInformation {
         self.get_tag::<BasicMemoryInfoTag, _>(TagType::BasicMeminfo)
     }
 
+    /// Search for the basic memory info tag, return a mutable reference.
+    pub fn basic_memory_info_tag_mut(&mut self) -> Option<&mut BasicMemoryInfoTag> {
+        self.get_tag_mut(TagType::BasicMeminfo)
+            .map(|tag| unsafe { &mut *(tag as *mut Tag as *mut BasicMemoryInfoTag) })
```
phip1611, line 255: I think `self.get_tag_mut(TagType::BasicMeminfo)` should be necessary?

multiboot2/src/memory_map.rs
```diff
@@ -280,6 +314,23 @@ impl EFIMemoryMapTag {
             phantom: PhantomData,
         }
     }
+
+    #[cfg(feature = "builder")]
```
phip1611, line 318:
```suggestion
```

phip1611 (changes requested):
fine with me. Users of this library could do this anyway with unsafe magic.. so just give them the necessary tools directly.


Please address the two open points. Otherwise, LGTM

YtvwlD:
Oh, yes, I missed that while rebasing.

phip1611 (approved):

YtvwlD:
Hm, I'm talking about UEFI and the default allocator: When I create boxes for the tags, they'll get put somewhere on the heap. This might be inside an existing memory region or it might allocate a new `LOADER_DATA` region.

The legacy map is only really accurate after calling `ExitBootServices`, I fear (after which the allocator doesn't work anymore).

Doing static allocations might work, but I'm not quite fond of that.

phip1611:
```rust
// SAFETY: BootInformation contains a const ptr to memory that is never mutated.
// Sending this pointer to other threads is sound.
unsafe impl Send for BootInformation {}
```

If we merge this, this needs to be removed. Another breaking change. I think this could be okay, as one can wrap the MBI in a (custom) Mutex..

YtvwlD:
Huh, I don't know? The struct does need to be mutable for these methods to be called, but I'm not sure that's enough.

phip1611:
I'm going to postpone this a bit and merge a few other things first. The rebase needed here will have a few conflicts - I'm sorry for that!

YtvwlD:
Sure, just ping me when I should rebase.

YtvwlD:
So, this should work (brb, testing this), but it really moves the `mut` through the whole codebase and I'm not happy about it.

YtvwlD:
I don't understand the miri failure, but this has better lifetimes and the integration test should work. I'm still not happy about having to touch so much code.

YtvwlD:
[Using `AsMut`](https://lab.whitequark.org/notes/2016-12-13/abstracting-over-mutability-in-rust/) might help with not-carrying-`mut`-through-the-whole-code, but I'm unsure whether this would look any better.

YtvwlD:
I think this looks better, though I'm not sure combining `EFIMemoryAreaIter` and `EFIMemoryAreaIterMut` is the way to go. (Combining `TagIter` and `TagIterMut` didn't work because there's pointers in them.)

YtvwlD:
That `unnecessary pub(self)` wasn't introduced by this pull request.

YtvwlD:
Oh, great!

(Don't mind me, this is just a rebase.)

YtvwlD:
(This is just a rebase.)

YtvwlD:
Yeah sure, take your time. I think this is the simplest change I can come up with right now, but if you find a simpler one, that'd be great.

phip1611:
Sorry. I'm closing this due to massive internal refactorings, necessary for memory safety and removing any UB.

However, I think that the new design (based in `multiboot2-common`) can act as base to allow mutable 
references to multiboot2 structures

YtvwlD:
Hm, I've been trying to port this over to the new design, but I'm currently stuck with quite a lot of duplications (`BytesRefMut`, `TagIterMut`, and about half of the functions in `DynSizedStructure`), so I'm not sure that this is the way to go.

phip1611:
I think having additional `ByresRefMut` and `TagIterMut` is fine and necessary. The duplication is fine.

I'm not sure how `structDynSizedStructure` and `trait MaybeDynSized` need to be altered. Perhaps we also need a `trait MaybeDynSizedMut` trait.

YtvwlD:
Thanks for the input and pointing me to the diagrams! They make the two stages of parsing clearer, but I didn't fully grasp the last one, yet.

In any case, this seems to be a bit harder than I thought; mostly because `TagIterMut` doesn't seem to be straightforward, so I might take some time to get it right.

(In the mean time, what follows is a partial rebase; please ignore it. :)

phip1611:
I'm sorry for the inconveniences. But all the refactorings were necessary to get rid of all UB.

PR #142, opened 2023-06-19
multiboot2: remove odd offset thingy

phip1611:
This removes the weird `load_with_offset` thingy which can also be done via an adjusted pointer. -> less complexity

Furthermore, the Multiboot2 information structure now captures the lifetime of the context where it is created.


phip1611:
Hey @YtvwlD 

What do you think about this API change? I'd like to get rid of the weird offset thingy. I do not see a reason why it is necessary at all. Please change my mind, if I miss something.

YtvwlD:
I did not understand why it was there, either. And the PR that added this came with no explanation either, right?

Perhaps this is for segmentation or something like that?

phip1611:
So, you are also in favor of this API change, @YtvwlD , right?

YtvwlD:
I don't have an opinion on this.

PR #149, opened 2023-06-20
fix memory issue in memory-map

phip1611:
Via git-bisecting, I could find that https://github.com/rust-osdev/multiboot2/commit/d198dceffe544b6324ba4c204270d9a1d934abba broke something with the (legacy) memory map tag. This MR fixes the problem and adds a unit test for it.

Also see: https://github.com/rust-osdev/multiboot2/pull/133


phip1611:
That was a fun one, @YtvwlD :D

PR #152, opened 2023-06-21
multiboot2: fix memory issue in boxed_dst_tag

phip1611:
This fixes multiple memory issues (wrong allocation and wrong fat pointer creation) that were introduced in #133 with 544e1ca555fb740eea272b936a41672eeef4d1ce.

I found the issues by running miri.


phip1611:
@YtvwlD FYI. There were two memory issues in your `boxed_dst_tag` function. Not blaming you, just FYI :)

phip1611:
Yeah, I thought so. It's a minefield, could have happened to me as well :)

PR #155, opened 2023-06-21
multiboot2: model MBI as DST

phip1611:
This is an improvement that enables better memory checks by miri. Furthermore, it is technically the right way to model this in Rust.

Hopefully, eventually, this will also simplify #128 #117 

This change is entirely internal, and it won't add more breaking changes to the next release.

### DIFF
--- a/multiboot2/src/builder/information.rs
+++ b/multiboot2/src/builder/information.rs
@@ -1,7 +1,7 @@
 //! Exports item [`InformationBuilder`].
 use crate::builder::traits::StructAsBytes;
 use crate::{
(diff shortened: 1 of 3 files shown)


phip1611:
Unfortunately, this still doesn't satisfy miri. `&*ptr` in `TagTrait::from_base_tag` is still causing problems. Miri reports 

>  trying to retag from <208001> for SharedReadOnly permission at alloc80640[0x8], but that tag does not exist in the borrow stack for this location - this error occurs as part of retag at alloc80640[0x0..0x10]

PR #156, opened 2023-06-22
builder: use new abstraction to guarantee alignment

phip1611:
Both builder structs (`HeaderBuilder`, `BootInformationBuilder`) return a `Vec<u8>`. However, there is no guarantee of an 8-byte alignment in that case. As long as the `allocator_api` is not stable, I chose the following workaround:

Use a regular `Vec<u8>` with a capacity of `expected_len + 7`. This way, I can add zeroes at the beginning to the vec. A wrapper type returns a slice to the actual and aligned bytes of the corresponding structure.


phip1611:
@YtvwlD I'm curious. Did you encountered problems with the builder in your project due to alignment issues?

If the allocator "is friendly", you might never have encountered any problems.

phip1611:
Ah, interesting. On Windows, the allocator triggered one of my new assertions

https://github.com/rust-osdev/multiboot2/actions/runs/5344793550/jobs/9689626355?pr=156

phip1611:
Strictly speaking, the Vec -> Box conversion is overkill. I could also keep the vector in the byte wrapper.

YtvwlD:
I did not encounter any errors yet. But you're right: According to the spec the struct should be 8-byte aligned.

PR #162, opened 2023-06-29
Make InformationBuilder adhere the API guidelines.

scholzp:
Rust [API guidelines](https://rust-lang.github.io/api-guidelines/type-safety.html#builders-enable-construction-of-complex-values-c-builder) state that methods of builders should return `self` to allow chaining.

Add missing documentation strings to builder functions.

### DIFF
--- a/multiboot2/src/builder/information.rs
+++ b/multiboot2/src/builder/information.rs
@@ -287,72 +287,126 @@ impl InformationBuilder {
         Self::build_add_bytes(bytes, &EndTag::default().struct_as_bytes(), true);
     }
 
-    pub fn basic_memory_info_tag(&mut self, basic_memory_info_tag: BasicMemoryInfoTag) {
-        self.basic_memory_info_tag = Some(basic_memory_info_tag)
+    /// Adds a 'basic memory information' tag (represented by [`BasicMemoryInfoTag`]) to the builder to later build the
+    /// boot information from.
+    pub fn basic_memory_info_tag(mut self, basic_memory_info_tag: BasicMemoryInfoTag) -> Self {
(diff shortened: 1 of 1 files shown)


multiboot2/src/builder/information.rs
```diff
@@ -287,72 +287,126 @@ impl InformationBuilder {
         Self::build_add_bytes(bytes, &EndTag::default().struct_as_bytes(), true);
     }
 
-    pub fn basic_memory_info_tag(&mut self, basic_memory_info_tag: BasicMemoryInfoTag) {
-        self.basic_memory_info_tag = Some(basic_memory_info_tag)
+    /// Adds a 'basic memory information' tag (represented by [`BasicMemoryInfoTag`]) to the builder to later build the
```
phip1611, line 290: nit: "to the builder to later build the boot information from"

I think this is obvious, as it is a builder :)

phip1611 (changes requested):
Please add a changelog entry

You can create a new 
```
## Unreleased
- foo

``` 

section in the changelog.

multiboot2/src/builder/information.rs
scholzp: You are totally right. :D
I removed the respective parts.

scholzp:
> Please add a changelog entry
> 
> You can create a new
> 
> ```
> ## Unreleased
> - foo
> ```
> 
> section in the changelog.

I added the respective changelog entry. Because the changes made it necessary to modify the tests, I think this qualifies as a breaking change.

phip1611 (approved):

phip1611:
Please fix the integration test

phip1611 (approved):

phip1611:
Thanks for your contribution!

PR #163, opened 2023-07-04
memory_map: support non-standard types

phip1611:
Support non-standard types in the memory map. We use this for example in our virtualization stack based on [Hedron](https://github.com/cyberus-technology/hedron).

Thanks to @scholzp for finding this issue.


phip1611:
Please let me know if this works for you, @scholzp

PR #170, opened 2023-07-14
multiboot2: builder: allow to add custom tags

### LINKED ISSUE #168: multiboot2: builder: Add ability to add custom tags
When I added the builder, I completely forgot about custom tags.

phip1611:
Allow the builder to use custom tags. The challenge here was that `TagTrait` is not object safe, so having `Vec<dyn TagTrait>` is not an option.

This closes #168.

## Steps to Undraft
The builder allows to set multiple SMBios Tags but the BootInformation only allows to get one.. which one is correct?

### DIFF
--- a/multiboot2/Changelog.md
+++ b/multiboot2/Changelog.md
@@ -1,5 +1,9 @@
 # CHANGELOG for crate `multiboot2`
 
+## 0.19.0 (2023-07-14)
+- `InformationBuilder` now also allows to add custom tags. The new public method
+  `add_tag` was introduced for that.
+
 ## 0.18.1 (2023-07-13)
 - Documentation improvements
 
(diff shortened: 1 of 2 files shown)


phip1611:
@YtvwlD Hey, could you help me please to figure out what's right? When you added the builder originally, you added the ability to add multiple SmBios tags. However, the boot information only allows to get a single SmBios tag. Which is correct?

YtvwlD:
I didn't see any limitation to a single tag in the spec. And I thought it would make sense to provide all available SMBIOS (major) versions.
So, from the perspective of a bootloader: If I get both SMBIOS 2 and SMBIOS 3 from the firmware, I can put each of them in a tag and pass both to the operating system.

phip1611:
I don't have a test setup to reproduce having SMBIOS tags. For simplicity, I restrict it to one tag for now for consistency. Otherwise, I'd need something like a `SMBIosTagIterator`.

I create a follow-up ticket tho.

PR #171, opened 2023-07-14
multiboot2: foundation for mutable reference to tags

phip1611:
This is the foundation for a mutable boot information and another approach than #140. By using the `NonNull` facility, we can ensure that the compiler will not optimize away any mutable operations. I could not find a way to use `UnsafeCell` here in a meaningful way. 

I think this approach has the advantage that the public API is barely changed - unlike in #140.

### DIFF
--- a/multiboot2/src/boot_loader_name.rs
+++ b/multiboot2/src/boot_loader_name.rs
@@ -39,7 +39,7 @@ impl BootLoaderNameTag {
     ///
     /// ```rust,no_run
     /// # use multiboot2::{BootInformation, BootInformationHeader};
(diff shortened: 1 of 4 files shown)


phip1611:
I'm still unsure but I think that this way, we might have more success. What do you think, @YtvwlD ?

YtvwlD:
oh, yes, this looks interesting (if it worked :upside_down_face:)

phip1611:
Obsolete as new API was too breaking. This needs a new approach. However, I think it can be nicely integrated in the new `multiboot2-common` lib

PR #195, opened 2023-12-12
updating header to not include multiboot with alloc

elbiazo:
Currently, when using multiboot2-header with no alloc (default-features = false), you wil get `error: no global memory allocator found but one is required` error.  Reason being , when including multiboot2 for MBI tags, it will also include alloc and builder. This patch fixes that so that you can use multiboot2-header as intended without alloc when it is off.


phip1611 (approved):

phip1611:
Thanks, good catch!

PR #225, opened 2024-08-14
Integration test uefi

phip1611:


phip1611:
I think this is not important enough - closing.

PR #232, opened 2024-09-10
Replaces dead link home page multiboot2-header package definition

filiplajszczak:
Looks like for package in a subdirectory of a repo there is no better way then linking to "tree/main"

The alternative would be to get rid of `homepage` key completely.


phip1611 (approved):
Thanks! Good catch

PR #290, opened 2026-03-20
Added elf library for elf code.

an-owl:
I needed to get the full ELF header I figured I'd implement #247 in the process. 

- In the front end everything is basically the same except I've added a method to get the whole section header.
- In the backend I've removed all the magic numbers and replaced them with constants from `::elf::abi`
- I've added a helper for the section header pointer, it handles offsetting and fetching the section header, it uses an intermediate type so callers can eliminate separate code paths.
- I removed the "strlen" code in favour of using `core::ffi::CStr` to handle it. Originally I wanted to use `CStr` to store the section name in `ElfSection` however it doesn't implement `Copy`
- `ElfSection::string_table` never needed to be `unsafe`, it is no longer `unsafe`
- ELF defines a program/user defined block of section header values which I've added.
- The `ElfSectionInner` trait is no longer needed because it now uses `elf::section::SectionHeader` for both ELF32 and ELF64. I've left it in solely for familiarity of others.  I can remove it if you want.
- `ElfSection` now **copies** the section header instead of maintaining a pointer to it. I can revert this behavior if you want.

PR #292, opened 2026-04-21
Rewrite: Add elf library for elf_sections.rs

an-owl:
This is a complete rewrite from #290, using `elf::sections::SectionHeaderTable` to construct the iterator. Unlike #290 however, I decided not to care about version compatibility. This removes or replaces a number of existing functions.

One change in particular is how to fetch the section name. The existing method in my opinion is flawed, if the ELF sections are relocated then it's not possible to get the name. I've changed this so the string table must be fetched separately, which can then be passed to a helper. This helper returns a `&core::ffi::CStr` instead of a `&str`, the ELF specification doesn't define the character encoding so we should not force it to be Unicode.

An important behavioral change that isn't obvious is that "unused" (`SHT_NULL`) sections are no longer skipped. This threw me for a loop during testing. If they're there it's probably there for a reason, so I think its a bad idea to skip it, especially if you can just use `.filter(_)`.

### DIFF
--- a/multiboot2/src/elf_sections.rs
+++ b/multiboot2/src/elf_sections.rs
@@ -176,6 +176,7 @@ impl ElfSectionExt for SectionHeader {
             elf::abi::SHT_DYNSYM => ElfSectionType::DynamicLoaderSymbolTable,
             elf::abi::SHT_LOOS..=elf::abi::SHT_HIOS => ElfSectionType::EnvironmentSpecific,
             elf::abi::SHT_LOPROC..=elf::abi::SHT_HIPROC => ElfSectionType::ProcessorSpecific,
+            elf::abi::SHT_LOUSER..=elf::abi::SHT_HIUSER => ElfSectionType::UserDefined,
             e => {
                 log::warn!("Unknown section type {e:x}. Treating as ElfSectionType::Unused");
                 ElfSectionType::Unused
@@ -251,6 +252,10 @@ pub enum ElfSectionType {
     /// Values in this inclusive range (`[0x7000_0000, 0x7FFF_FFFF)`) are
     /// reserved for processor-specific semantics.
     ProcessorSpecific = elf::abi::SHT_LOPROC,
+
+    /// Values in this inclusive range (`[0x8000_0000, 0x8FFF_FFFF)`) are
+    /// reserved for user-specific semantics.
+    UserDefined = elf::abi::SHT_LOUSER,
 }
 
 bitflags! {
--- a/multiboot2/src/elf_sections.rs
+++ b/multiboot2/src/elf_sections.rs
@@ -1,6 +1,7 @@
 //! Module for [`ElfSectionsTag`].
 
 use crate::{TagHeader, TagType};
+use core::ffi::{CStr, FromBytesUntilNulError};
 use core::fmt::{Debug, Formatter};
 use core::mem;
 use elf::endian::NativeEndian;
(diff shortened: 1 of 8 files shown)


an-owl:
Any progress on this?

phip1611:
Hey! thanks for pinging me! Will look into this soon - I'll put it on my list

multiboot2/src/elf_sections.rs
```diff
@@ -81,6 +81,12 @@ impl ElfSectionsTag {
     }
 }
 
+impl<'a> From<&'a ElfSectionsTag> for SectionHeaderTable<'a, NativeEndian> {
```
phip1611, line 109: I'm unsure about the value-add of this. It feels like a misuse of `From` - could you please add a motivation to the commit?

multiboot2/src/elf_sections.rs
```diff
@@ -117,61 +115,20 @@ impl Debug for ElfSectionsTag {
...
-    flags: u32,
-    addr: u32,
-    offset: u32,
-    size: u32,
-    link: u32,
-    info: u32,
-    addralign: u32,
-    entry_size: u32,
-}
+/// Extension trait for [SectionHeader] containing getters for rust-native types
```
phip1611: ``/// Extension trait for [`SectionHeader`] containing getters for rust-native types``

multiboot2/src/elf_sections.rs
```diff
@@ -117,61 +115,20 @@ impl Debug for ElfSectionsTag {
...
+pub trait ElfSectionExt {
```
phip1611, line 146: There is only one impl for this trait. You do this to extend the type `SectionHeader` which comes from the `elf` library, yes?

multiboot2/src/elf_sections.rs
```diff
@@ -60,6 +61,32 @@ impl ElfSectionsTag {
...
+        // Info for this here
+        // https://docs.oracle.com/cd/E23824_01/html/819-0690/chapter6-43405.html @ `e_shstrndx`
+        let strtab_index = match self.shndx as u16 {
+            elf::abi::SHN_UNDEF => return None,
+            elf::abi::SHN_XINDEX => shdr_table.get(0).unwrap().sh_link as usize,
+            i => i as usize,
+        };
+
+        let strtab_hdr = shdr_table.get(strtab_index).ok()?;
+        // todo: Should this check that `strtab_hdr.sh_type == elf::abi::SHT_STRTAB`?
```
phip1611, line 78: what about this todo?

multiboot2/src/elf_sections.rs
```diff
@@ -60,6 +61,32 @@ impl ElfSectionsTag {
...
+
+        // SAFETY: The multiboot2 spec defines that sections are always loaded at `sh_addr`.
+        // Casting through `usize` will not truncate data on 32bit systems because the multiboot2 loads all sections below u32::MAX
+        Some(unsafe {
+            core::slice::from_raw_parts(
+                strtab_hdr.sh_addr as usize as *const u8,
```
phip1611: I think this should be: 

`core::ptr::with_exposed_provenance(strtab_hdr.sh_addr as usize)`

which is the more future-proof API

phip1611 (changes requested):
Lovely - **thanks**! This is a great contribution! Only a few nits. Also, please update the changelog.

multiboot2/src/elf_sections.rs
an-owl: Fixed in e455889

multiboot2/src/elf_sections.rs
an-owl: Fixed in 9b87c7a

multiboot2/src/elf_sections.rs
an-owl: That's the idea, otherwise you need to cast the section type & flags through raw pointers. It also adds a method to fetch the section name, `elf` can only fetch the string table from `elf::elfBytes` or `elf::ElfStream` which obviously we don't have.

multiboot2/src/elf_sections.rs
an-owl: The idea is that `ElfSectionTag` is effectively

```
struct ElfSectionTag {
    multiboot_data: ...,
    header_table: SectionHeaderTable,
}
```
So it would be nice to implement `AsRef` so it can be passed to functions taking `impl AsRef<SectionHeaderTable>`. But this causes a use-after-free because this `SectionHeaderTable::new()` does not return a reference.

multiboot2/src/elf_sections.rs
an-owl: I have absolutely no strong feelings one way or another. 

[elf does this](https://github.com/cole14/rust-elf/blob/5702acc510d616aff43bfe37b0d0764d2cba899a/src/elf_bytes.rs#L475-L480), But I'm not sure if we should bother checking. I feel it's the bootloaders responsibility to check this.

phip1611: fine with me, we can leave this comment there for now

multiboot2/src/elf_sections.rs
phip1611: okay - thanks!

phip1611:
`Merge remote-tracking branch 'origin/elf' into elf`

You have a weird git history. Please clean up your git history without any `merge` commits and rebase

an-owl:
I really need to learn to use git correctly, but I think I fixed it.

multiboot2/src/elf_sections.rs
```diff
@@ -39,16 +44,21 @@ impl ElfSectionsTag {
...
+    const fn class(&self) -> elf::file::Class {
+        use elf::section::{Elf32_Shdr, Elf64_Shdr};
+        const SHDR_ELF32_SIZE: usize = size_of::<Elf32_Shdr>();
+        const SHDR_ELF64_SIZE: usize = size_of::<Elf64_Shdr>();
+
+        match self.entry_size as usize {
+            SHDR_ELF32_SIZE => elf::file::Class::ELF32,
+            SHDR_ELF64_SIZE => elf::file::Class::ELF64,
+            _ => {
+                panic!("Unknown ELF section entry size");
```
phip1611, line 59: I'd prfer to return an error rather than panic here - but we can do that in a follow-up. No need to delay this PR any longer

phip1611 (approved):
Thanks for your contribution and your patience! Nice!

PR #299, opened 2026-06-15
elf: remove From impl

phip1611:
This impl doesn't feel very idiomatic and a direct invocation is clearer.

Follow-up of #292.

### DIFF
--- a/multiboot2/src/elf_sections.rs
+++ b/multiboot2/src/elf_sections.rs
@@ -64,7 +64,7 @@ impl ElfSectionsTag {
     /// Returns the string table data, if it's present.
     #[must_use]
     pub fn string_table(&self) -> Option<&[u8]> {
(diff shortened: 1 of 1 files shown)


multiboot2/src/elf_sections.rs
```diff
@@ -64,7 +64,7 @@ impl ElfSectionsTag {
     /// Returns the string table data, if it's present.
     #[must_use]
     pub fn string_table(&self) -> Option<&[u8]> {
-        let shdr_table: SectionHeaderTable<NativeEndian> = self.into();
+        let shdr_table = SectionHeaderTable::new(NativeEndian, self.class(), &self.sections);
```
phip1611, line 67: FYI @an-owl - I think the From impl was not very idiomatic Rust. I decided to switch back to a normal constructor.

Nevertheless: thanks for your contribution!

